PR #3, opened 2020-02-21
Feature/short proposal period

n13:


n13:
Deploy instructions

- Deploy settings
- Call reset action on settings
- Deploy proposals
- Call onperiod on proposals next start of a moon cycle

PR #12, opened 2020-03-02
Feature/scheduler

IanMendozaJaimes:
- Scheduler modified to execute actions from harvest, proposals, referendums and forum contracts. It could execute any action of any contract with the permission @execute and if the action does not need parameters.
- Removed deferred transactions from calcplanted, calctrx, calcrep from harvest to prevent them to auto-schedule.
- Schedulers now takes seconds as timestamp.
- Added the cancelexec action to cancel the execute deferred transaction if necessary.

### DIFF
--- a/src/seeds.scheduler.cpp
+++ b/src/seeds.scheduler.cpp
@@ -7,11 +7,21 @@
 
 bool scheduler::isRdyToExec(name operation){
     auto itr = operations.find(operation.value);
-    check(itr != operations.end(), "The operation does not exist.");
+
+    if (itr == operations.end()) {
+        return false;
+    }
+    if(itr -> pause > 0) {
+        print("transaction " + operation.to_string() + " is paused");
+        return false;
+    }
+
     uint64_t timestamp = eosio::current_time_point().sec_since_epoch();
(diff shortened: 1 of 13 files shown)


src/seeds.scheduler.cpp
```diff
@@ -93,29 +165,50 @@ ACTION scheduler::execute() {
...
-                //permission_level(get_self(), "scheduled"_n),
-                permission_level{get_self(), "active"_n},
+                permission_level{itr -> contract, "execute"_n},
                 itr -> contract,
                 itr -> operation,
                 std::make_tuple()
             );
 
-            a.send();
+            transaction txa;
```
n13: @IanMendozaJaimes @7flash So this is sending a delayed transaction for each scheduler call with a delay of 0?

I also see that every time this executes, it will only execute 1 scheduled operation. That's good because then we have enough CPU. But I think we need to then iterate over it sorted by last execution date (timestamp) so we make sure we never forget any operations - if we always go by the op name, then it could be that we always execute the first op, but the other ops never even get looked at.

I'll make this change

n13:
Closing this it has all been merged

PR #22, opened 2020-03-18
Cancel proposal action

IanMendozaJaimes:
- Implemented the cancel proposal action
- Implemented refund action
- Validate only the proposal's owner can stake

### DIFF
--- a/src/seeds.proposals.cpp
+++ b/src/seeds.proposals.cpp
@@ -252,6 +252,7 @@ void proposals::stake(name from, name to, asset quantity, string memo) {
 
       auto pitr = props.find(id);
       check(pitr != props.end(), "no proposal");
+      check(from == pitr->creator, "only the creator can stake into a proposal");
 
       props.modify(pitr, _self, [&](auto& proposal) {
           proposal.staked += quantity;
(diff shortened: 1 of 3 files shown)


src/seeds.proposals.cpp
```diff
@@ -212,6 +212,28 @@ void proposals::update(uint64_t id, string title, string summary, string descrip
...
+}
+
+void proposals::refund(uint64_t id) {
+  auto pitr = props.find(id);
+
+  check(pitr != props.end(), "Proposal not found");
+  require_auth(pitr->creator);
+  check(pitr->status == name("cancel"), "Proposal state must be cancel");
+
+  withdraw(pitr->creator, pitr->staked, contracts::bank);
```
n13, line 234: This needs to delete the proposal from the table 

Otherwise I can call cancel multiple times, and receive a refund multiple times.

Alternatively, set "staked" to 0.

test/proposals.test.js
```diff
@@ -195,6 +212,20 @@ describe('Proposals', async assert => {
     expected: 0
   })
 
+  assert({
+    given: 'stake from not owner',
+    should: 'has error',
+    actual: notOwnerStake,
```
n13, line 218: add unit test for calling refund multiple times

src/seeds.proposals.cpp
n13: Ok I am implementing this now, I'll delete from table

n13:
deployed

PR #28, opened 2020-03-25
Feature voice decay and rollover

IanMendozaJaimes:
- Implemented voice decay
- Created rollover table
- Modified permissions to allow scheduler to execute voicedecay


7flash:
Can you explain voice decay for reference here please?

n13:
Voice decay is too complicated IMO

Propose to simply cut 13% of voice daily

PR #71, opened 2020-07-01
Feature/incremental count

n13:
Incremental counting for rep and cbs scores.

### DIFF
--- a/src/seeds.accounts.cpp
+++ b/src/seeds.accounts.cpp
@@ -264,6 +264,7 @@ void accounts::refreward(name account, name new_status) {
       item.account = referrer;
       item.community_building_score = community_building_points.value;
     });
+    size_change("cbs.sz"_n, 1);
   }
 
   // see if referrer is org or individual (or nobody)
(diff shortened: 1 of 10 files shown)


src/seeds.accounts.cpp
```diff
@@ -542,19 +584,228 @@ void accounts::genesis(name user) // Remove this after Golive
...
+
+  uint64_t total = sizes.get("rep.sz"_n.value, "user rep size not set").size;
+  uint64_t current = chunk * chunksize;
+  auto rep_by_rep = rep.get_index<"byrep"_n>();
+  auto ritr = start_val == 0 ? rep_by_rep.begin() : rep_by_rep.lower_bound(start_val);
+  uint64_t count = 0;
+
+  while (ritr != rep_by_rep.end() && count < chunksize) {
+
+    uint64_t rank = (current * 100) / total;
```
7flash, line 646: what is this formula?

src/seeds.accounts.cpp
```diff
@@ -542,19 +584,228 @@ void accounts::genesis(name user) // Remove this after Golive
...
-      });
+  // DEBUG REMOVE THIS - THIS IS SO THE FUNCTION DOESN"T RUN AMOK
+  //if (chunk > 10) { 
+  //  return;
+  //}
+
+  uint64_t total = sizes.get("rep.sz"_n.value, "user rep size not set").size;
+  uint64_t current = chunk * chunksize;
+  auto rep_by_rep = rep.get_index<"byrep"_n>();
+  auto ritr = start_val == 0 ? rep_by_rep.begin() : rep_by_rep.lower_bound(start_val);
```
7flash, line 641: so we are sorting only this chunk - how do we make sure it's sorted across the chunks?

src/seeds.accounts.cpp
n13: same ranking formula as we used before - using current index / total to get a RANK from 0-99

n13: This formula has not changed, and is also tested in the unit tests

src/seeds.accounts.cpp
n13: It's sorted by the index - it's always sorted by the system. We only walk over the sorted array and calculate the rank based on the current index vs size of array

Issue #59, opened 2020-05-12
Vote Neutral


n13:
Add a new vote: neutral. Neutral is the same as favour or against, but gives no points and costs no votes.

It creates an entry in the votes table, however. 

This is so we can give reputation to those who voted for all

Voting neutral 
- casts no points
- uses no trust tokens
- counts as having voted as reputation gain is concerned
- counts as quorum

n13:
Finished

PR #74, opened 2020-07-07
Issue #59, Vote Neutral

IanMendozaJaimes:
- Added a neutral vote. 
- Deleted boolean variable `favour` from `vote_table` and added name variable `vote` to store the `trust`, `distrust` and `neutral` options.
- Modified `favour` and `against` functions to work along with `neutral` in a single function.

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -42,12 +42,17 @@ CONTRACT proposals : public contract {
 
       ACTION against(name user, uint64_t id, uint64_t amount);
 
+      ACTION neutral(name user, uint64_t id);
+
       ACTION onperiod();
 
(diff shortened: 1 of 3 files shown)


include/seeds.proposals.hpp
```diff
@@ -111,7 +117,8 @@ CONTRACT proposals : public contract {
           uint64_t proposal_id;
           name account;
           uint64_t amount;
-          bool favour;
```
n13, line 114: Question: Will this impact existing stored votes tables? Will they become unreadable?

Is there a solution to migrate old votes tables over, or would it be better to simply use a new type name?

@IanMendozaJaimes

n13:
superceded by another PR

Issue #51, opened 2020-05-10
Transaction Limit By Planted Seeds


n13:
## Limit transactions by planted

[1] planted seed = N tx / day where [N] is a setting

N = 10 initially, so we get 10 tx / day / planted seed

Implement this for 2x for now compared to constitution - we want to prevent abuse of the system, but we also want to prevent users running into this.

Larger of 7 / week OR 2x planted value.

## Implementation hints

- Minimum 1 / day (0 planted)
- We reset the transactions weekly
- incremental reset with delayed transactions so reset doesn't fail
- Only affects individuals, no check for organizations or bioregions
- look into using existing function check_limit 
- add resetweekly in token contract to scheduler so it's called every day 
- add new setting

<img width="549" alt="image" src="https://user-images.githubusercontent.com/65412/81511879-cb748580-931c-11ea-9f9d-53ea799e24d2.png">

IanMendozaJaimes:
1. I see check_limit is commented, is this issue replacing the check_limit's functionality?
2. Which quantity are you referring to by 2x compared to constitution? Is it the planted value?, in that case, is the planted value the amount of planted seeds?
3. Sorry but I think I don't understand what you mean by this line: Larger of 7 / week OR 2x planted value.
4. If we configure the resetweekly function to be called daily, isn't going to reach a point where transactions will be reset daily instead of weekly? 
5. The resetweekly function does not handle batches, do I have to implement this incremental reset as well right?

n13:
# Deployment

## Testnet

- [ ] token
- [ ] scheduler


## Mainnet 

- [ ] token
- [ ] scheduler

n13:
deployd

n13:
Deployed

PR #86, opened 2020-07-31
Issue Transaction Limit By Planted Seeds #51

IanMendozaJaimes:
- Added a validation in transfer action from token to limit the number of transactions per user
- Added settings txlimit.mul and txlimit.min
- Added reset weekly to the scheduler
- Added permission to token

### DIFF
--- a/src/seeds.token.cpp
+++ b/src/seeds.token.cpp
@@ -228,32 +228,31 @@ void token::reset_weekly_aux(uint64_t begin) {
 
   auto batch_size = config.get(name("batchsize").value, "The batchsize parameter has not been initialized yet.");
   auto sym_code_raw = seeds_symbol.code().raw();
-  uint64_t counter = 0;
+  uint64_t count = 0;
 
   transaction_tables transactions(get_self(), sym_code_raw);
 
-  auto titr = transactions.begin();
-  std::advance(titr, begin);
-  while (titr != transactions.end() && counter < batch_size.value) {
-    transactions.modify(titr, get_self(), [&](auto& user) {
+  auto titr = begin == 0 ? transactions.begin() : transactions.lower_bound(begin);
+  while (titr != transactions.end() && count < batch_size.value) {
+    transactions.modify(titr, _self, [&](auto& user) {
       user.incoming_transactions = 0;
       user.outgoing_transactions = 0;
       user.total_transactions = 0;
       user.transactions_volume = asset(0, seeds_symbol);
     });
     titr++;
-    counter++;
+    count++;
   }
 
-  if (counter == batch_size.value) {
+  if (titr != transactions.end()) {
     transaction trx{};
     trx.actions.emplace_back(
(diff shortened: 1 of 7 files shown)


src/seeds.token.cpp
```diff
@@ -193,33 +222,51 @@ void token::check_limit(const name& from) {
...
+
+  auto batch_size = config.get(name("batchsize").value, "The batchsize parameter has not been initialized yet.");
   auto sym_code_raw = seeds_symbol.code().raw();
+  uint64_t counter = 0;
 
   transaction_tables transactions(get_self(), sym_code_raw);
 
   auto titr = transactions.begin();
-  while (titr != transactions.end()) {
+  std::advance(titr, begin);
```
n13, line 236: This will not work since "advance" has linear time, it's the same as iterating through all accounts

You need to use the same methods we are using in scores refactoring where the begin value is the actual account name value (not the counter)

This is in the branch https://github.com/JoinSEEDS/seeds-smart-contracts/pull/84

n13: Sample implementation here in rankrep - it's always the same pattern, we avoid all counting - advance, next, all these are linear time and will fail with > 1000 table entries.

We only use find and get on the table, which are logN time.

```
void accounts::rankreps() {
  rankrep(0, 0, 200);
}

void accounts::rankrep(uint64_t start_val, uint64_t chunk, uint64_t chunksize) {
  require_auth(_self);

  // DEBUG REMOVE THIS - THIS IS SO THE FUNCTION DOESN"T RUN AMOK
  //if (chunk > 10) { 
  //  return;
  //}

  uint64_t total = get_size("rep.sz"_n);
  uint64_t current = chunk * chunksize;
  auto rep_by_rep = rep.get_index<"byrep"_n>();
  auto ritr = start_val == 0 ? rep_by_rep.begin() : rep_by_rep.lower_bound(start_val);
  uint64_t count = 0;

  while (ritr != rep_by_rep.end() && count < chunksize) {

    uint64_t rank = (current * 100) / total;

    rep_by_rep.modify(ritr, _self, [&](auto& item) {
      item.rank = rank;
    });

    current++;
    count++;
    ritr++;
  }

  if (ritr == rep_by_rep.end()) {
    // Done.
  } else {
    // recursive call
    uint64_t next_value = ritr->by_rep();
    action next_execution(
        permission_level{get_self(), "active"_n},
        get_self(),
        "rankrep"_n,
        std::make_tuple(next_value, chunk + 1, chunksize)
    );

    transaction tx;
    tx.delay_sec = 1;
    tx.send(next_value + 1, _self);
    
  }

}
```

IanMendozaJaimes: Hello, I committed the change. I removed the `advance` and I used `lower_bound` instead as you suggested.

n13: Perfect

n13:
#51

Issue #81, opened 2020-07-27
App tracking


n13:
## Apps can call a special function with account / app that tracks app use

In organization contract:
- org can register an app (name, string longname)
- org contract track app (account, appname) - signed by account
- banapp(name) with contract rights - ban misbehaving apps

For now we can just track the number of app uses on chain.

We can rely on manual checking to see that the app dodn't cheat and just call trackapp without doing anything - it's only expected to be preloaded on existing actions, like votes, transfers, etc

But I think there's no way to verify that on chain, so we can rely on off chain checking for the time being, I think we'll only have Seeds Passport and Seeds Light Wallet for now anyway, we can test this system, and add more sophisticated tracking later.

It would be nice to track DAU rather than just actions too

IanMendozaJaimes:
1. By App you mean Seeds Passport and Seeds Light Wallet?
2. Do we have to track the usage per user or per action maybe? 
3. When org registers an app, it means we have to create a new table to store the apps register by this org?, like: 
   app_table { 
       name app_name;
       string app_long_name;
       boolean is_banned;
       uint64_t  number_of_uses;
   } scoped by org
4. Does the number of app uses have to be stored in a table?
5. How could I track when a user uses the app?
6. In the documentation I saw DAU is the same as App points, but what are App points? I didn't see them in the documentation

n13:
Thinking about DAU I think it can be done with a users table scoped by app and a counter - the counter increases when a user uses the app and the counter is reset every 24h, and when that happens the daily active users for the app are stored in another table with date and DAU also scoped by app

That's one possible solution - it avoids having to empty the tables all the time

IanMendozaJaimes:
For DAU tracking, instead of 2 tables, we could use a single table with the following structure, what do you think?

dau_table {
  uint64_t id;
  name account;
  uint64_t date // day could be set up to be year-month-dayT00:00:00
  uint64_t number_app_uses;
} scoped by app

with this table structure we would create a new entry per user every day or update the entry if there is one, account and date would be secondary indices

IanMendozaJaimes:
Hello I have another doubt:
- If org can register apps, who has to sign this actions, the contract itself or some member belonging to the organization?.

n13:
Good question. Let’s make it the owners of the org for now ? 

Not any member. Only the owner.

n13:
It should also cost Seeds. But I can add that later

n13:
Done

PR #87, opened 2020-07-31
App tracking #81

IanMendozaJaimes:
- Added functions: registerapp, banapp, appuse, cleandaus and cleandau
- Added app_table, dau_table and dau_history_table

### DIFF
--- a/src/seeds.organization.cpp
+++ b/src/seeds.organization.cpp
@@ -318,8 +318,9 @@ ACTION organization::subregen(name organization, name account) {
     vote(organization, account, -1 * getregenp(account));
 }
 
-ACTION organization::registerapp(name organization, name appname, string applongname) {
-    require_auth(get_self());
+ACTION organization::registerapp(name owner, name organization, name appname, string applongname) {
+    require_auth(owner);
+    check_owner(organization, owner);
 
     auto orgitr = organizations.find(organization.value);
     check(orgitr != organizations.end(), "This organization does not exist.");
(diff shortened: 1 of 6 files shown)


src/seeds.organization.cpp
```diff
@@ -307,4 +328,136 @@ ACTION organization::subregen(name organization, name account) {
...
+        transaction tx;
+        tx.actions.emplace_back(clean_dau_action);
+        tx.delay_sec = 1;
+        tx.send((appitr -> app_name).value + 1, _self);
+
+        appitr++;
+    }
+}
+
+ACTION organization::cleandau (name appname, uint64_t timestamp, uint64_t start) {
```
n13, line 419: Is cleandaus supposed to be called by scheduler?

IanMendozaJaimes: Yes `cleandaus` is the function that should be called daily, and then `cleandaus` will call `cleandau` for each app.

I committed a change that stores the current day and then the function `cleandaus` can be called at any time and it will just store the information of the previous days, just in case the `cleandaus` funtion fails for some reason it could be called at any time. Also when using `appuse` action, it also check the day, in case of any error from `cleandaus`, `appuse` action will store the old info for that user so the information is consistent. 

I added the `cleandaus` action to the scheduler, I configured it to be called every 12 hours just in case.

n13: Cool!

n13 (approved):

Issue #61, opened 2020-05-12
Funding Fee and Referendum fee burning


n13:
## Spec 

Props and referendums cost 5% of the requested amount but limited to 555 and 11,111 

Fee returned of referendum passes, burned if not

pp 47 of constitution

IanMendozaJaimes:
Hello, I have some questions:
- I would like to clarify if I understood well. For props, we have to check if the stake is 5% of the prop quantity or if it is within the given limits (555, 11,111), if the stake meets one of those conditions then the prop is ready to be voted, is that correct? In case it is correct, I noticed referendums table doesn't have a quantity field, how should I determine the requested amount?

n13:
@IanMendozaJaimes Correct referendums don't have amounts, so leave the fees there as they are.

Referendum will then keep fixed cost as it is, and have the burn / return policy - I think it already has that.

Thanks for noticing!

n13:
Max was set to 1,111, changed to 11,111

n13:
Done - needs to be deployed, waiting for App to update.

PR #94, opened 2020-08-12
Funding Fee and Referendum fee burning #61

IanMendozaJaimes:
- Added parameters: `propmaxstake` and `propstakeper`
- Added validations to: `vote_aux`, `onperiod` and `stake` 

## Deployment

- [ ] Deploy Settings
- [ ] Deploy Proposals when App is ready

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -53,6 +53,8 @@ CONTRACT proposals : public contract {
 
       ACTION decayvoice();
 
+      ACTION checkstake(uint64_t prop_id);
+
   private:
       symbol seeds_symbol = symbol("SEEDS", 4);
       name trust = "trust"_n;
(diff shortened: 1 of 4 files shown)


n13:
This is for #61

n13:
I had to change is_enough_stake as it wasn't doing as intended

The idea was the fee is 5% of the proposal funding request, but at minimum 555, and at maximum 11,111 

So it defaults to 5%, but is capped.

I also made a change where proposals that do not have enough stake are not moved to "active" - this is less confusing for clients as nobody can vote on these anyway.

n13:
Since staking is now rather complicated, I am considering adding a field so people know how much they have to stake

n13:
Added minstake table for proposals. Otherwise clients will be overwhelmed trying to figure this out.

n13:
@IanMendozaJaimes I made a few changes please check and comment here

IanMendozaJaimes:
Hello @n13, I checked the changes and everything seems great. 

The only thing we notice is that in this line:
https://github.com/JoinSEEDS/seeds-smart-contracts/blob/feature/funding-fee-burning/src/seeds.proposals.cpp#L353
the validation for the proposal stake is not necessary anymore since it is being checked now within `onperiod` action

Is there anything else to be done for this issue?

n13:
@IanMendozaJaimes I also added a minstake table as I feel it's going to be too error prone and complicated to know what the value is for apps. Just a service.

Issue #77, opened 2020-07-12
Fix harvest unit test


n13:
./scripts/seeds.js run harvest

Something has broken in the unt test

==:
+  (node:20816) UnhandledPromiseRejectionWarning: {"code":500,"message":"Internal Service Error","error":{"code":3050003,"name":"eosio_assert_message_exception","what":"eosio_assert_message assertion failure","details":[{"message":"assertion failure with message: cannot pass end iterator to modify","file":"wasm_interface.cpp","line_number":964,"method":"eosio_assert"},{"message":"pending console output: ","file":"apply_context.cpp","line_number":113,"method":"exec_one"}]}}
(node:20816) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)

PR #95, opened 2020-08-13
Fix harvest test #77

IanMendozaJaimes:
The error wasn't the one shown in the issue, but the test still failed. However it just failed because transactions needed some time to execute so I added some sleeps. 

I also had to comment this line in my nodeos as the comment suggests to run the tests:
https://github.com/JoinSEEDS/seeds-smart-contracts/blob/master/src/seeds.history.cpp#L102

I also noticed that the key variable used to identify the deferred transaction would be the same for two users with the same number of recorded transactions and thus if those two users make a transaction at the same time one of the transactions will fail. This could be really hard to happen in production but it was notorious in the tests so I just wanted to mentioned it.
https://github.com/JoinSEEDS/seeds-smart-contracts/blob/master/src/seeds.history.cpp#L103

### DIFF
--- a/test/harvest.test.js
+++ b/test/harvest.test.js
@@ -4,6 +4,10 @@ const { equals } = require("ramda")
 
 const { accounts, harvest, token, firstuser, seconduser, thirduser, bank, settings, history, fourthuser } = names
 
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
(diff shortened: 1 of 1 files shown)


n13:
Thanks @IanMendozaJaimes I noticed the same bug yesterday and deployed a hotfix. Chances are that's why some of the unit tests were failing

Issue #60, opened 2020-05-12
High, medium, and low thresholds


n13:
## Spec

High, medium and low thresholds for quorum and majority (unity)

-> pp 46 constitution

While we are at it, also add a description to each setting

## Implementation

Add the following fields to config
- impact high/med/low name
- description

All existing entries have "high" impact

IanMendozaJaimes:
Hello, I have some questions:
- Do we have to add a `description` field to the `config_table` from settings?
- Just to clarify, we have to add impact high/med/low as new parameters in `config_table`, right?
- To determine the proposal's impact, should we add the `impact` field in the `proposals` table? or is it outside this issue's scope?

n13:
> Hello, I have some questions:
> 
> * Do we have to add a `description` field to the `config_table` from settings?

Yes.

> * Just to clarify, we have to add impact high/med/low as new parameters in `config_table`, right?

Yes.

We also need a migration strategy

Normally I would make 2 branches, branch 1 is migration branch and deploys the new table structure with a new name - e.g. config2, and has a migrate action that migrates all data from config -> config2

But in our case since we initialize all settings in the reset action anyway, we can just deploy a migrate action that clears all data.

Then in branch 2 has the new data structure

So in this case - migrate branch has action "migrate" which just deletes the table

feature branch has new data structure and all gets initialized in init

Question: Since many contracts depend on config_table - do we need to recompile and redeploy them all too?

> * To determine the proposal's impact, should we add the `impact` field in the `proposals` table? or is it outside this issue's scope?

No, we don't need to modify proposals contract - proposals can figure out the impact and therefore the required votes % from config - that's the idea. 

In a proposal a user votes on a settings parameter - we can get the impact from the config table in the settings contract.

PR #98, opened 2020-08-14
High, medium, and low thresholds #60

IanMendozaJaimes:
- Added action `confwithdesc`. This action is called inside `reset` action and configures a parameter with a description. We added this action to not affect referendums contract as it is using `configure` action.

- I also created another branch called migrate/clean-settings: https://github.com/JoinSEEDS/seeds-smart-contracts/tree/migrate/clean-settings. In this branch there is an action in settings contract called `migrate` which only deletes the `config` table.

> Question: Since many contracts depend on config_table - do we need to recompile and redeploy them all too?
>
No, you don't need to neither recompile nor redeploy them as long as they don't need to access to the new `description` parameter.

I also noticed that these contracts didn't pass all their tests successfully:
- lending
- organization
- exchange
- token
it was not caused by the modifications in settings since they failed before I applied any change to my node:

### DIFF
--- a/include/seeds.settings.hpp
+++ b/include/seeds.settings.hpp
@@ -21,10 +21,14 @@ CONTRACT settings : public contract {
 
       ACTION setcontract(name contract, name account);
 
+      ACTION adddescriptn(name param, string description);
+
   private:
       TABLE config_table {
         name param;
         uint64_t value;
+        string description;
+
         uint64_t primary_key()const { return param.value; }
       };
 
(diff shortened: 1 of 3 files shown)


n13:
@IanMendozaJaimes thanks for the bug report I will look into failing tests and make a ticket

Still missing is the "impact" field as described in #60 

it's type name and for now all existing settings are set to "high"

IanMendozaJaimes:
Hello @n13, I added the changes requested, sorry for missing the field.

n13:
Deployed

Issue #114, opened 2020-09-25
Simple Vote Depreciation


n13:
A much simplified system

At half cycle and every day thereafter, figure out half cycle time, add up to 100%, remove a number of % each day

Example
cycle time = 2 weeks 
half cycle time = 1 week
At 1 week, every day deprecate votes by (100 / 7) =~ 14%

IanMendozaJaimes:
Hello, I have some questions:
1. Does the depreciation occur over the `voice table` or over the votes a user has given to a proposal?
2. Just to make sure, each day after the half of the cycle, the 100% is the current amount right?, I mean, let's suppose a user has  100 points, then after the first depreciation the user will have 100 - 14 = 86, then for the second day our 100% will be 86, thus the depreciation will be 86 - 12.04 = 73.96
3. For the decimals, should I use the four digits of precision in the integer or just round it?

n13:
> Hello, I have some questions:
> 
> 1. Does the depreciation occur over the `voice table` or over the votes a user has given to a proposal?


Voice table 

The purpose is to encourage users to vote early. Votes they have already cast remain unaffected 

So this only affects voice that you did not spend. 

Then people will want to vote before half cycle so they can vote with full power. 

> 2. Just to make sure, each day after the half of the cycle, the 100% is the current amount right?, I mean, let's suppose a user has  100 points, then after the first depreciation the user will have 100 - 14 = 86, then for the second day our 100% will be 86, thus the depreciation will be 86 - 12.04 = 73.96

Yes that way we don’t need to keep track of anything... just multiply by .85 

(Should be 15% not 14 as discussed on call I’ll change the spec too)

> 3. For the decimals, should I use the four digits of precision in the integer or just round it?

Rounding down would be appropriate. So they always lose at least 1 

So we got for floor( val * 0.85 )

n13:
We can make it 15% now and deal with it later when we have full moon cycles

PR #117, opened 2020-09-28
Feature/simple vote depreciation - #114

IanMendozaJaimes:
- Added `decayvoices` and `decayvoice` actions
- Added new settings
- Added to scheduler reset

## Deploy Steps

- [ ] Deploy proposals contract
- [ ] Deploy scheduler contract
- [ ] Deploy settings contract
- [ ] run updatePermissions
- [ ] run reset on settings contract

Amazingly, the voice decay timestamp was already in there so no migration of tables - foresight level visionary ;)

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -52,13 +52,15 @@ CONTRACT proposals : public contract {
 
       ACTION onperiod();
 
-      ACTION decayvoice();
-
       ACTION updatevoices();
 
       ACTION updatevoice(uint64_t start);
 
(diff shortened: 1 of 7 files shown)


n13 (changes requested):
Need to figure out
- How to set up scheduler

Need to add permissions for scheduler to be able to run

n13:
@IanMendozaJaimes 

So I think this will just work, if we add it to the scheduler and it fires every 24h - done. 

I changed the delay to 1/2 proposal cycle

PR #123, opened 2020-09-30
upgraded security on all incoming transfer actions

n13:
standardized checks on incoming assets

added get_first_receiver() check on all incoming actions

passed all unit tests.


n13:
deployed

PR #133, opened 2020-10-18
maximum ask of new accounts is 250,000

n13:
## Spec

- The first time a user creates a proposal, they cannot ask for more than 50,000
- If their project requires more funding, we will ask them to make a second proposal after their first one has passed and is in review.

## Implementation

- keep track of people's ask and success history
- For people with non-positive track record, max is 50,000

Still missing:

- [x] Unit test
- [x] Change value to 250k? 
- [x] Implementation: A failed proposal "resets" a user's score. We add up successes and failures - maybe not needed. 
- [ ] Add ability to blacklist?
- [x] Migrate old accounts into the trust table- so all previous proposals count. 

## Deployment

- [ ] deploy proposals and settings
- [ ] call reset on settings
- [ ] call migrathistry() after deploying

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -69,6 +69,10 @@ CONTRACT proposals : public contract {
       name abstain = "abstain"_n;
       name max_initial_ask_prop = "prop.maxask"_n;
 
+      name type_alliance = "alliance"_n;
+      name type_campaign = "campaign"_n;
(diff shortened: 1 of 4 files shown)


juliolrmonteiro:
There is a discussion for if this should be deployed or not.

n13:
Closing this since we may not want this.

Issue #129, opened 2020-10-16
Proposals Changes - Flexible Quorum


n13:
Yup! Copied above.
.90 / # of proposals = quorum required.

Min 7% max 20%
Just for now and see how this goes...

IanMendozaJaimes:
Hello, I have question:
Does any of these parameters have to be a setting?

n13:
Quorum is very different overall actually

We take the total number voted FOR and compare it with the number of votes for a specific proposal. 

![image](https://user-images.githubusercontent.com/65412/97449148-a9ee7100-1931-11eb-88ca-377c43659b03.jpeg)

n13:
Use total number of votes not just positive votes

PR #142, opened 2020-10-22
Proposal changes - flexible quorum #129

IanMendozaJaimes:
- Added new function to calculate the quorum
- Added size table
- Added tests

#129 

## Deployment

- [ ] deploy settings
- [ ] deploy proposals
- [ ] call initsz
- [ ] call initnumprop
- [ ] call reset settings

### DIFF
--- a/src/seeds.proposals.cpp
+++ b/src/seeds.proposals.cpp
@@ -212,7 +212,7 @@ void proposals::onperiod() {
     uint64_t current_cycle = c.propcycle;
 
     while (pitr != props.end()) {
-      if (pitr->stage == name("active")) {
+      if (pitr->stage == stage_active) {
         number_active_proposals += 1;
         votes_tables votes(get_self(), pitr->id);
         uint64_t voters_number = distance(votes.begin(), votes.end());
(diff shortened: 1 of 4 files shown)


src/seeds.proposals.cpp
```diff
@@ -89,16 +107,37 @@ uint64_t proposals::get_size(name id) {
...
     require_auth(_self);
 
     auto pitr = props.begin();
 
     auto prop_majority = config.get(name("propmajority").value, "The propmajority parameter has not been initialized yet.");
-    uint64_t quorum = config.find(name("propquorum").value)->value;
 
     uint64_t number_active_proposals = 0;
-    uint64_t total_eligible_voters = get_size("active.sz"_n);
+    uint64_t total_eligible_voters = get_size("active.sz"_n, contracts::accounts);
```
n13, line 139: active.sz should be in this contract's sizes as well - no need to leave than in the accounts sizes table?

IanMendozaJaimes: Yes, you are right, there is no need to user the account's sizes table. I left only the proposals' sizes table.
Just one doubt, does it means the accounts' sizes table should no be updated? I mean in the active.sz

n13: I changed it in master branch to that active size simply tracks voice table size. 

Only users that can vote are in the voice table. When someone gets demoted from citizen to resident, they are also removed from the voice table. So it's a lot simpler and more accurate to just keep track of voice. 

@IanMendozaJaimes do you agree? Or did I miss something?

n13 (approved):

Issue #130, opened 2020-10-16
Proposals Changes - Different Stakes for Alliances and Campaigns


n13:
Campaigns: 2.5% stake with 75,000 cap

Alliances: 1% stake and 15,000 cap.

## Implementation - 

Alliance and campaigns use the same stake caps at the moment
This will change it so they use different stake caps - need to introduce 2 new settings for this

IanMendozaJaimes:
Hello, I have some questions:
- The 2.5% and 1% are based on the percentage of Seeds required, right?
- Is the 75,000/15,000 cap the maximum amount of Seeds an alliance/campaign proposal can stake?
- Just to make sure, the percentages aren't settings as well, right?

PR #143, opened 2020-10-22
Proposals Changes - Different Stakes for Alliances and Campaigns #130

IanMendozaJaimes:
- Added new limits to the stakes for Campaigns and Alliances
- Added tests

#130

### DIFF
--- a/src/seeds.proposals.cpp
+++ b/src/seeds.proposals.cpp
@@ -20,6 +20,7 @@ void proposals::reset() {
   auto voiceitr = voice.begin();
   while (voiceitr != voice.end()) {
     voiceitr = voice.erase(voiceitr);
+    size_change("active.sz"_n, -1);
   }
 
   auto paitr = participants.begin();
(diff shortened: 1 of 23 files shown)


src/seeds.proposals.cpp
```diff
@@ -38,34 +38,61 @@ void proposals::reset() {
...
+uint64_t proposals::min_stake(asset quantity, name fund) {
+
+  double prop_percentage;
+  uint64_t prop_min;
+  uint64_t prop_max;
+
+  prop_min = config.get(name("propminstake").value, "The propminstake has not been initialized yet.").value;
+  
+  if (fund == bankaccts::campaigns) {
+    prop_percentage = 2.5;
```
n13, line 67: yes the % need to be settings too

n13: We could make them x 10,000 as integers, so 4 digits of precision like an asset. 

So 2.5% would be 25000 => 25,000 / 10,000 = 2.5 %

n13: @IanMendozaJaimes

n13: used to be called 
  confwithdesc(name("propstakeper"), 5, "Proposal funding fee in % - 5%", high_impact);

just need another one of those...

n13 (changes requested):

n13 (approved):

n13 (approved):

Issue #137, opened 2020-10-20
Separate voice distributions for Alliance vs Campaign type proposals


n13:
scope voice

leave main contract scope for campaign

make new scope for alliance

distinguish proposals

helper function
name get_type(proposal->fund) -> returns prop type as it is currently determined by prop type

On a vote, determine which pool it comes out of by type

on vote distribution, distribute to both pools

IanMendozaJaimes:
Hello I have some questions:
> name get_type(proposal->fund) -> returns prop type as it is currently determined by prop type,
- What is the prop type?, currently I think it is determined by the type of fund, but there is no attribute called prop type, what would be the difference?
- Currently I think the banks are normal accounts, and I couldn't find a way to distinguish that some user is a bank, well other than the hardcoded name, is it planned to have several bank accounts of each type?, if so, is there a table to identify the banks and their types?
> On a vote, determine which pool it comes out of by type
- Sorry, but I think I don't understand what you mean by pool. The vote table is currently scoped by proposal id, but if the proposal scope changes, then different proposals may have the same id. Is this what you mean?, What could be a good solution to that problem?, is it fine to separate the 64 bits id in two? I mean one part for the prop type and the other the proposal id? these could help, but it could be a migration problem if there are proposals already.

n13:
> Hello I have some questions:
> 
> > name get_type(proposal->fund) -> returns prop type as it is currently determined by prop type,
> 
> * What is the prop type?, currently I think it is determined by the type of fund, but there is no attribute called prop type, what would be the difference?

Yeah we currently decide by the fund selected. 

That's why I was sugesting the helper method - a bit over-specified - do it any way you see fit please ;) 

> * Currently I think the banks are normal accounts, and I couldn't find a way to distinguish that some user is a bank, well other than the hardcoded name, is it planned to have several bank accounts of each type?, if so, is there a table to identify the banks and their types?

We use the hard-coded banks for the time being, until we need more flexibility

Eventually all these hard coded banks will be drained and the proposals will come out of the available harvest in this cycle - then this will have to be revisited. 

It would be nicer to specify them for example as settings, so we can change them but realistically they will not change until we hit go live. They're not a long term thing anyway. 

> 
> > On a vote, determine which pool it comes out of by type
> 
[...]

PR #144, opened 2020-10-27
Separate voice distributions for Alliance vs Campaign type proposals #137

IanMendozaJaimes:
- Added new scope for alliance proposals
- Added new methods to modify voices
- Added tests

The action `migratevoice` is there only to take all the existing voice entries and copy them into the new scope. But the fact that a user does not exist in the voice table scoped by `alliance` does not affect the rest of the methods. Indeed some methods can add the missing voice entry for a user that already has voice scoped by `contract's name`.

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -75,11 +75,18 @@ CONTRACT proposals : public contract {
 
       ACTION testvdecay(uint64_t timestamp);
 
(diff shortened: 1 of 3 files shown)


n13 (approved):

n13 (approved):

n13:
Looks good. Getting a little complex with the different scopes and edge cases.

Issue #18, opened 2020-03-11
Proposals: Evaluation Phase


n13:
# Specification

## Evaluation Phase

- After a proposal is approved, it can be voted down by all those who voted for it - that is every yes vote can change into a no vote during the evaluation phase
- Partial payouts: In the first cycle, 10% of total is paid out. Over the total number of cycles the proposal is running, each cycle a number of additional funds is paid out. Either in a linear or in a stepping up fashion. 
- Minimum run time is 3 cycles
- Default payout is 10% (initial), 30% (end of cycle 1), 30% (end of cycle 2), 30% (end of cycle 3)
 
### Stepped payout

We could also use a stepping function where each cycle the number paid out increases, i.e.

1. 10%
2. 20%
3. 30%
4. 40%

### New state: evalulate
Instead of "passed" / "closed", proposal goes into "evalulate" state

### onperiod changes
In onperiod, we check all eval state proposals
- if the votes changed, it is closed down
- if the votes didn't change (still passing) we pay out the next chunk

### Data changes and create action changes

In the creation of a proposal, and the props table, there's additional parameters
- Initial payout (default 10%, can't be higher than 20%)
- number of cycles (minimum 3)
- props table: add "cycle this proposal was passed" so we know which cycle we are in, compared to total cycle, which we're already counting.
- Alternatively, add "age" and increase age every time on onperiod, until it reached number of cycles. Age seems more intuitive and less error prone here. 

Add a function that calculates the number of Seeds to pay out in a given cycle, given the following parameters:
total cycle, cycle proposal was passed, total number of seeds requested

juliolrmonteiro:
Discussion today with Nik:
- Move the proposal data structure to a JSON-like(map<string><Any>) data prop.
- We need to validate the schema for the map, based on the version (Json schema like validation)
- This will allow the evolution of Proposals, to evolve using Roles etc.

IanMendozaJaimes:
Hello I have some questions:
- If a user voted no, then that user can't change their vote later on in the next proposal's cycles, right?
- Is there an upper limit on the number of cycles a proposal can run?


>Either in a linear or in a stepping up fashion
- Does this have to be chosen when the proposal is created?


>Instead of "passed" / "closed", proposal goes into "evalulate" state
[...]

PR #146, opened 2020-10-29
Feature/evaluation phase #18

IanMendozaJaimes:
## #18 

- Added evaluation phase to proposals
- Added tests
- Added new fields to proposal table

About the Max's Document mentioned in the last meeting, we think this new table structure will suppose a great change in the way the information is stored, and as it is still pretty new, it might have some issues not yet discovered. So that's why I only added the extra fields to the proposals table.


n13 (approved):
All looks great but would love to have a migrate action ...

n13:
Deployed

PR #147, opened 2020-11-05
Adjust referral rewards

juliolrmonteiro:


n13:
Looks good

src/seeds.accounts.cpp
```diff
@@ -425,10 +426,24 @@ void accounts::update(name user, name type, string nickname, string image, strin
...
+  auto rem_balance = acc.balance;
+  if (quantity > rem_balance) {
+    // Should not fail in case not enough balance
+    // check(false, ("DEBUG: not enough balance on "+bankaccts::referrals.to_string()+" = "+rem_balance.to_string()+", required= "+quantity.to_string()).c_str());
+    return;
+  }
 
-  // TODO: Check balance - if the balance runs out, the rewards run out too.
+  // Checks the current SEEDS price from tlosto.seeds table
+  auto firstprice = pricehistory.rbegin()->seeds_usd;
```
n13, line 440: this is called "firstprice" but it is actually the last price entry, no?

src/seeds.accounts.cpp
```diff
@@ -425,10 +426,24 @@ void accounts::update(name user, name type, string nickname, string image, strin
...
+  auto lastprice = pricehistory.begin()->seeds_usd;
```
n13, line 441: this is called lastprice but is actually the first price

n13 (changes requested):
just name change of variable - everything else looks good

src/seeds.accounts.cpp
juliolrmonteiro: Done, changed it to lastprice

src/seeds.accounts.cpp
juliolrmonteiro: Done, changed it to initialprice

juliolrmonteiro:
Also fixed all the tests, so that if "exchange.reset()" is uncommented and working the tests pass here.

n13 (approved):

n13 (approved):

PR #148, opened 2020-11-06
limit rank to 99

n13:
## Analysis

This happens when rankings / points change while the rank is in progress. 

For example, user A has 5 points, rank runs and sets user A's rank to 0, the ranking ends at position 200

User A receives another 20 points from somewhere

Ranking continues at position 200, but now user A is located at position 450 (for example) and will eventually be ranked again - maybe at rank 15 for example. 

That rank is correct - that's fine - but we are counting up on how many users we have ranked to determine rank, and so the count is now 1 too many, and when we get to the end of the table, count == size, and rank is set to 100.

The rank will even out again on the next run of the ranking.

The same thing will happen in many other ways, but they're less visible - e.g. the rank for someone will be too high or too low, but it evens out the next run when they're once again ranked correctly.

##      The Good Solution - Table Locks
      
The reason this is not atomic is that we go over the table in chunks, so inbetween chunks the actual ranking can change.
      
The correct way to solve this would be to wait until the ranking is finished, and to apply changes in reputation only once that has happened. Like a manually implemented table lock. (rep or any other number we are ranking)
      
For now that's too complex - a lock would have to be between changing reputation and ranking. 
Ranking could simply wait until there's no lock but applying reputation we would have to store changes in a secondary table rep_delta, apply all changes there while rep ranking is happening and apply the delta back to the full table once rep ranking is finished.
 
And this has to be done for all 8 or so rankings we are doing. The rep delta table has to be cleared while iterating and could exceed 200 items, so it also has to be run in chunks.

### DIFF
--- a/include/seeds.forum.hpp
+++ b/include/seeds.forum.hpp
@@ -5,6 +5,7 @@
 #include <tables/user_table.hpp>
 #include <tables/config_table.hpp>
 #include <tables/size_table.hpp>
+#include <utils.hpp>
 
 using namespace eosio;
 using std::string;
--- a/include/utils.hpp
+++ b/include/utils.hpp
@@ -6,7 +6,6 @@
 using namespace eosio;
 
 namespace utils {
-
   const uint64_t seconds_per_day = 86400;
   const uint64_t seconds_per_minute = 60;
   const uint64_t seconds_per_hour = seconds_per_minute * 60;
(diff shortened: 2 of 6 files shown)


n13:
Please review and especially if you can think of other ways to implement a good solution let me know!

Main issue we are ranking a list that is changing while we're ranking it.

juliolrmonteiro:
Could we have a "dirty" flag on users at the ranking table, to indicate they have already been ranked (clear), and once they receive more points they become "dirty" again, so If the algorithm goes by a user that has the dirty flag on it removes that user from the ranked table before adding them again on the proper place?

Now I am going to review the implemented solution.

n13:
I think dirty clear would just shift the issue to that flag ... my feeling. It might simply cause other edge cases to not work 

Things may be added to or removed from the list too but my main argument without thinking through all the details is that you can’t substitute a table lock - mutual exclusive access - with anything else... not for the general case. 

It’s possible that the flag would make errors less likely though. Nothing would be double counted. But then some other mechanism has to go over the table and set everything to dirty before the next run and then we have the same issue again some might get skipped and not set to dirty...

n13:
"Living Ranking"

Issue #138, opened 2020-10-20
Harvest - Restructure Transactions Table in history


n13:
- [ ] 3 new Tx tables and a migrate function and 128 bit index on other 

## Why

For the harvest protocol we need to calculate QEV - qualifying economic volume

So we need to find transactions that quality

## Problem
Efficiently finding the top 26 transactions in a period

## Solution
Combining "other" and "volume" figures into a single index

IanMendozaJaimes:
Hello, I have some questions:
- Could you describe the 3 new tables you mentioned?
- Could you spec a little bit more what a qualified transaction is?
- Why is the "other" needed in the new index? is it important to know who made the transaction to determine if the transaction is in the top 26?

juliolrmonteiro:
- a table scoped by the individual account sending the transaction, sorted by "to", plus volume (destination)
- a table scoped by the organization transactions, indexed by "from" plus "volume"
indexed by name and volume
- migrate the current transaction into these 2 new tables

juliolrmonteiro:
a qualified transaction is: the largest 26 transactions between 2 parties, capped at 1777 SEEDS

IanMendozaJaimes:
Hello, I have some questions:

- Just to make sure I understood, the QEV is the sum of the amount of seeds of the top 26 transactions in a given period, but those 26 transactions are chosen from all transactions processed during the given amount of time, and the transactions could belong to any user, right? 

- Another thing related, the 26 transactions, should belong to 26 different users?, or it is possible to have 26 transactions made by only one user.

- I think I have some confusion with `by_to_quantity` index. 

It is because if we have an index of name + volume, and let's say we have users A, B and C. Let's suppose we are exploring the transactions made by user A, so, for instance if A has transferred to B and C, when we explore the table using `by_to_quantity` index in reverse order, we will get the highest transaction A has made to C, but this is not necessarily the highest transaction the user A has made, and this is not necessarily a transaction made within the period we are looking for. So I still need to traverse the entire transaction table for A.

If I understood wrong, could you please explain me again what the QEV is and the usage of the intex `by_to_quantity`?

[...]

PR #164, opened 2020-11-12
Harvest - Restructure Transactions Table in history #138

IanMendozaJaimes:
- Added migration tables `transaction_table_migration` and `org_tx_table_migration`
- Added migration functions to transfer the information from the original tables to the migration tables

The original tables with the new index and the function to transfer back the information are in the branch feature/harvest-resctructure-transactions-table


n13:
We changed everything here with the UQEV calculations - closing

Issue #140, opened 2020-10-21
Calculate hourly mint quantity


n13:
Calculate daily QEV growth and outgoing hourly harvest

## Spec

Take the tables calculated in #139  with the daily values, use them to calculate how many seeds we need to grow

- Compare volume today with volume 1 quarter ago
- This is the increase in QEV - QEV growth
- Take circulating supply back then, grow it by QEV growth -> Target supply
- Calculate target  - Current circ supply  -> delta 
- spread delta over (29.5 * 24) hours = 708 hours, divide -> number to distribute each hour going forward
- Store this number - mint rate
- Refresh the number 24h later

IanMendozaJaimes:
Hello, I have some questions:
- By the volume from 1 quarter ago, do you mean literally just consider the volumes from the months 3,6,9 and 12, or consider the volume three months ago, for instance if the current month is 4 (April), the previous QEV would be the QEV from month 1 (January)?
- Where should I store the QEV growth and the mint rate in a new table?
- Is QEV growth = current_QEV - previous_QEV ? that means QEV growth is measured in SEEDS right?
- mint rate = delta / 708,  right?

juliolrmonteiro:
By the volume from 1 quarter ago, do you mean literally just consider the volumes from the months 3,6,9 and 12, or consider the volume three months ago, for instance if the current month is 4 (April), the previous QEV would be the QEV from month 1 (January)? --> Second one, rolling average QEV

juliolrmonteiro:
Where should I store the QEV growth and the mint rate in a new table?
--> harvest contract can have a new table to store this variation of the QEV and mint rate

juliolrmonteiro:
Is QEV growth = current_QEV - previous_QEV ? that means QEV growth is measured in SEEDS right?
--> It's a percentage growth

juliolrmonteiro:
Refer to https://docs.google.com/document/d/1C4w9Ol8VGabCIcQDVPDrwcTRoJXBqhrb7VjslwQbUGU/edit# page 93

juliolrmonteiro:
The minting should happen hourly, based on the daily calculation of delta QEV

Issue #139, opened 2020-10-21
Harvest Calculate Monthly QEV


n13:
## Spec

Keep a table updated every 24h with 
- QEV (running over the trailing 29.5 days)
- circulating supply
- planted
- burned (?? not sure that's feasible - we don't keep track of that so far?)

## Questions
- Think through whether storing circulating supply is sufficient
- define what circulating supply is
- store extra data if we need to (burned?)

## Implementation

- Calculation - crawls tx history set up in #138 
- Store in table

- [ ] Calculate monthly QEV function with scheduler. Calc every 24h. Store circulating supply and planted and burned as well. Goes back 150 days

IanMendozaJaimes:
Hello, I have some questions:
- By using the 3 tables mentioned in issue 138 would be possible to track the circulating supply, planted seeds and burned, right?
- Could you spec a bit more about what the  circulating supply is?
- I think I don't understand what you mean by "Goes back 150 days"
- Just to make sure I understood, this table will be updated daily, and the data to take into account is the data between the current day and  29.5 days ago
- Is it convenient to create a new entry in this table each cycle? I think this will be needed in the issue 140

juliolrmonteiro:
The circulating supply is already there (updatecirc, on token.cpp)

juliolrmonteiro:
We should keep the last 150 snapshots of the circulating supply in a table

juliolrmonteiro:
Just to make sure I understood, this table will be updated daily, and the data to take into account is the data between the current day and 29.5 days ago ---> Yes

juliolrmonteiro:
Store daily calculations on a table, and consider only 2 qualifying transactions(between 2 unique accounts, limit to 1777) per day. From that the monthly QEV can be calculated naturally

PR #165, opened 2020-11-12
Harvest Calculate Monthly QEV #139

IanMendozaJaimes:
- Added functions to calculate QEV and table `qev_table`
- This pull request also includes the issue #140 as it requires all the changes made in issue #139 
- This pull request assumes the changes in transaction tables are already implemented (issue #138 )

### DIFF
--- a/include/seeds.harvest.hpp
+++ b/include/seeds.harvest.hpp
@@ -33,10 +33,12 @@ CONTRACT harvest : public contract {
         sizes(receiver, receiver.value),
         total(receiver, receiver.value),
         harveststat(receiver, receiver.value),
+        qevgrowths(receiver, receiver.value),
         config(contracts::settings, contracts::settings.value),
         users(contracts::accounts, contracts::accounts.value),
         rep(contracts::accounts, contracts::accounts.value),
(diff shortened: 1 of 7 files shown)


n13:
Ok... I wanted to also change the structure so that ortanizations are inside the main transaction tables .... that would make it easier to iterate over all accounts.

Then we could just go over all users, and look up all their tx scores by volume with the name scope

Otherwise we have to filter out duplicates from the orgs table, becasue it contains incoming and outgoing

Originally I thought it would be nicer if it had incoming and outgoing. But the problem is we then potentially count things twice as all the org incoming transactions are someone else's outgoing ones, and already counted as outgoing

For that reason I thought we could just store an org outgoing the same exact way as individual outgoing - then we only have one table

And store org incoming in another org specific table, which is then only used to calculate the org's score, bot for nothing else. 

IDK if that makes sense. Basically figuringout what gets stored where so we can do 3 different calculations:

- Total QEV - only looks at outgoing
- Org tx score - looks at specific org incoming and outgoing
- Individual tx score - looks at outgoing only

n13:
I think this has been superceded by the unified QEV calculation - @IanMendozaJaimes please confirm

Issue #167, opened 2020-11-15
Unified QEV Calculation


juliolrmonteiro:
Have a single transactions history table stored in a daily scope

So here's a completely new idea on how to do all our transaction point and QEV calculations in one table

The key differences to the old system are:
- daily tables and daily summaries
- limit of 2 per day for transactions between the same accounts (replaces 26 per month)
- Everything is clocked daily and stored in a table scoped by day (in UTC 00:00)

[Google sheet with new table structure](https://docs.google.com/spreadsheets/d/1-Q6bkm7qFXwoden_P7pQV6QM5XY8afrLADS_aqaG9bo/edit?usp=sharing)

|ID                                         |From|To                         |Volume|Qualifying Volume|FromPoints|ToPoints|Time|
|-------------------------------------------|----|---------------------------|------|-----------------|----------|--------|----|
|0                                          |a   |b                          |30    |30               |30        |5       |....|
|1                                          |a   |b                          |200   |200              |60        |10      |....|
|3                                          |c   |d                          |60,000|1777             |40        |33      |....|
|4                                          |a   |b                          |400   |0                |0         |0       |....|

Transaction comes in, is added to the list - checks whether or not it's more than 2 between 2 accounts and does the calculation for qualifying vol, from points, to points

If there's already 2 transactions and this is the 3rd, it's not added to the list - simplification we don't take the 2 highest volume transactions, we just take the first 2

[2] is a parameter

As a transaction comes in we 
- generate the scope from the timestamp
- save it to the table
- place a delayed transaction with the id

#### The delayed transaction (id, timestamp)
calculates from points, to points, QEV - this all depends on reputation scores / individual vs. org (only orgs get points as the receover and so on. and enters these points into tables for individual tx points, org tx points, and QEV

These are daily tables that accumulate a day with the timestamp, scoped by name

So each user has a 120 day trailing table of time stamped tx points - their total tx points can then be easily calculated by just adding them up

scope user

|Timestamp UTC 00:00  |points|
|------------------------|------|
|0                                       |3  
|1                                       |24  

QEV
|Timestamp UTC 00:00  |Qualifying volume|
|------------------------|------------------|
|0                                      |220
|1                                       |701  

## tx score for orgs
Just like we do now transaction score for orgs is calculated in harvest, but it can just add up all the points in the account's table

## tx score for individuals
same as for orgs - the calculation can run as usual, but it will just sum up the last X days

## QEV
[...]

PR #173, opened 2020-11-26
Unified QEV Calculation #167

IanMendozaJaimes:
- Added new tables to store daily transactions and total points per day. Added a totals table
- Added QEV calculation
- Modified transaction points calculation for orgs and users
- Modified number of transactions validation for becoming a resident/citizen user or reputable/regenerative org
- Added tests

I also pushed a branch called migration/unified-qev-calculation, there I put a method called `migrateusers`, that action will copy the transactions for users and orgs and store them in the new tables using the new criteria to count transactions. The method doesn't clean the old transactions tables just in case.

## Deployment

- [x] ./scripts/seeds.js run accounts harvest history organization settings

- [x] Test accounts
- [x] Test harvest
- [x] Test history
- [x] Test organization

## Deploy Testnet
- [ ] Deploy 4 contracts to testnet
- [ ] ./scripts/seeds.js updatePermissions
- [ ] reset scheduler
- [ ] Run migration on testnet
- [ ] Ensure all values are as expected

## Deploy Mainnet
- [ ] Deploy 4 contracts
- [ ] ./scripts/seeds.js updatePermissions
- [ ] reset scheduler
- [ ] Run migration
- [ ] Ensure all values are as expected

### DIFF
--- a/src/seeds.history.cpp
+++ b/src/seeds.history.cpp
@@ -12,12 +12,6 @@ void history::reset(name account) {
   while(hitr != history.end()) {
     hitr = history.erase(hitr);
   }
-  
-  // org_tx_tables orgtx(get_self(), account.value);
-  // auto oitr = orgtx.begin();
-  // while (oitr != orgtx.end()) {
-  //   oitr = orgtx.erase(oitr);
-  // }
 
   transaction_points_tables transactions(get_self(), account.value);
   auto titr = transactions.begin();
@@ -143,8 +137,6 @@ void history::trxentry(name from, name to, asset quantity) {
 
   double trx_multiplier = std::min(max_transaction_points, quantity.amount) / 10000.0;
 
-  print("MIRA: to = ", to, ", multiplier = ", utils::get_rep_multiplier(to), ", volume = ", trx_multiplier, "\n");
-  
   transactions.emplace(_self, [&](auto & transaction){
     transaction.id = transaction_id;
     transaction.from = from;
@@ -242,8 +234,6 @@ void history::savepoints(uint64_t id, uint64_t timestamp) {
       current_itr = ft_itr;
     }
 
-    print("count = ", count, ", volume = ", ft_itr -> volume, "\n");
-
     ft_itr++;
     count++;
   }
@@ -260,11 +250,37 @@ void history::savepoints(uint64_t id, uint64_t timestamp) {
     transactions_by_from_to.erase(current_itr);
   }
 
+  save_from_metrics (from, from_points, qualifying_volume, day);
+
+  if (uitr_to -> type == name("organisation")) {
+    transaction_points_tables trx_points_to(get_self(), to.value);
+    auto trx_itr_to = trx_points_to.find(day);
+
+    if (trx_itr_to != trx_points_to.end()) {
+      trx_points_to.modify(trx_itr_to, _self, [&](auto & item){
+        item.points += to_points;
+      });
+    } else {
+      trx_points_to.emplace(_self, [&](auto & item){
+        item.timestamp = day;
+        item.points = to_points;
+      });
+    }
+  }
+
+  if (uitr_from -> type != name("organisation")) {
+    send_update_txpoints(from);
+  }
+}
+
+void history::save_from_metrics (name from, int64_t & from_points, int64_t & qualifying_volume, uint64_t & day) {
   transaction_points_tables trx_points_from(get_self(), from.value);
   qev_tables qevs(get_self(), from.value);
+  qev_tables qevs_total(get_self(), get_self().value);
 
   auto trx_itr = trx_points_from.find(day);
   auto qev_itr = qevs.find(day);
+  auto qev_total_itr = qevs_total.find(day);
 
   if (trx_itr != trx_points_from.end()) {
     trx_points_from.modify(trx_itr, _self, [&](auto & item){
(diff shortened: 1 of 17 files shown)


n13 (approved):

juliolrmonteiro (approved):
Looks good to me, I would just try to merge tables `qev_table` and `transaction_points_table` in a single one, but this is not required.

src/seeds.history.cpp
```diff
@@ -113,202 +126,236 @@ void history::trxentry(name from, name to, asset quantity) {
...
+  double to_trx_multiplier = std::min(max_transaction_points_organizations, quantity.amount) / 10000.0;
+
+  transactions.emplace(_self, [&](auto & transaction){
+    transaction.id = transaction_id;
+    transaction.from = from;
+    transaction.to = to;
+    transaction.volume = quantity.amount;
+    transaction.qualifying_volume = std::min(transactions_cap, quantity.amount);
+    transaction.from_points = uint64_t(ceil(from_trx_multiplier * utils::get_rep_multiplier(to)));
+    transaction.to_points = uint64_t(ceil(to_trx_multiplier * utils::get_rep_multiplier(from)));
```
n13, line 157: I have a question on this
- if "to" is an individual then transaction.to_points  should be == 0 (as individuals do not receive points on incoming transactions)

src/seeds.history.cpp
```diff
@@ -113,202 +126,236 @@ void history::trxentry(name from, name to, asset quantity) {
...
   
-  if (from_is_organization) {
-    org_tx_tables orgtx(get_self(), from.value);
-    orgtx.emplace(_self, [&](auto& item) {
-      item.id = orgtx.available_primary_key();
-      item.other = to;
-      item.in = false;
-      item.quantity = quantity;
-      item.timestamp = eosio::current_time_point().sec_since_epoch();
+  double to_trx_multiplier = std::min(max_transaction_points_organizations, quantity.amount) / 10000.0;
```
n13, line 148: Why is this divided by 10000?

Should this not be called "capped_amount" - seems like that's what it is the quantity, capped to the limit ?

src/seeds.history.cpp
```diff
@@ -113,202 +126,236 @@ void history::trxentry(name from, name to, asset quantity) {
...
-    return;
+  auto ft_itr = transactions_by_from_to.find(from_to_id);
+  auto current_itr = ft_itr;
+
+  while (ft_itr != transactions_by_from_to.end() && 
+      count <= max_number_transactions && 
+      ft_itr -> from == from && ft_itr -> to == to) {
+
+    if (ft_itr -> volume < current_itr -> volume) {
+      current_itr = ft_itr;
```
n13, line 242: current_itr is set to the smallest of the - potentially - 3 transactions

src/seeds.history.cpp
```diff
@@ -113,202 +126,236 @@ void history::trxentry(name from, name to, asset quantity) {
...
+  int64_t from_points = int64_t(titr -> from_points);
+  int64_t to_points = int64_t(titr -> to_points);
+  int64_t qualifying_volume = int64_t(titr -> qualifying_volume);
 
-  auto transactions_by_other = orgtx.get_index<"byother"_n>();
-  auto tx_other_itr = start_val == 0 ? transactions_by_other.begin() : transactions_by_other.lower_bound(start_val);
+  if (count > max_number_transactions) {
+    from_points -= current_itr -> from_points;
+    to_points -= current_itr -> to_points;
+    qualifying_volume -= current_itr -> qualifying_volume;
```
n13, line 256: So this could either end up in some positive values, or it could be all 0 if the new transaction is the smallest transaction

Then it deletes the smallest, so ends back up with 2 transactions

src/seeds.history.cpp
```diff
@@ -113,202 +126,236 @@ void history::trxentry(name from, name to, asset quantity) {
...
 
-  name      current_other = name("");
-  uint64_t  current_num = 0;
-  double    current_rep_multiplier = 0.0;
+    transactions_by_from_to.erase(current_itr);
+  }
 
-  //print("orgtxpt for "+org.to_string());
+  save_from_metrics (from, from_points, qualifying_volume, day);
```
n13, line 262: so save_from_metrics gets *additional* points - which could potentially be 0, but cannot be negative? Right?

test/harvest.test.js
```diff
@@ -504,45 +528,48 @@ describe("harvest transaction score", async assert => {
...
+  // console.log("transfer with 10 rep, 2 accounts have rep")
   await contracts.accounts.testsetrs(thirduser, 75, { authorization: `${accounts}@active` })
+  await transfer(seconduser, thirduser, 10, '0'+memoprefix)
+
   await checkScores([10, 16], [0, 50], "2 reputation, 2 tx", "0, 50 score")
 
   let expectedScore = 15 + 25 * (1 * 1.5) // 52.5
   console.log("More than 26 transactions. Expected tx points: "+ expectedScore)
-  for(let i = 0; i < 40; i++) {
+  for(let i = 0; i < 1; i++) {
```
n13, line 551: This is weird code - i = 0 and < 1?? That's not an iteration ;)

test/harvest.test.js
```diff
@@ -504,45 +528,48 @@ describe("harvest transaction score", async assert => {
...
-  console.log("transfer with 10 rep, 2 accounts have rep")
-  await contracts.token.transfer(seconduser, thirduser, '10.0000 SEEDS', '0'+memoprefix, { authorization: `${seconduser}@active` })
+  // console.log("transfer with 10 rep, 2 accounts have rep")
   await contracts.accounts.testsetrs(thirduser, 75, { authorization: `${accounts}@active` })
+  await transfer(seconduser, thirduser, 10, '0'+memoprefix)
+
   await checkScores([10, 16], [0, 50], "2 reputation, 2 tx", "0, 50 score")
 
   let expectedScore = 15 + 25 * (1 * 1.5) // 52.5
   console.log("More than 26 transactions. Expected tx points: "+ expectedScore)
```
n13, line 550: comment needs to change

src/seeds.history.cpp
IanMendozaJaimes: Oh yeah, that's right I am not setting them to 0, although the field "to" is not being used to calculate transaction points for individuals, but yeah I will change it.

src/seeds.history.cpp
IanMendozaJaimes: I divided it by 10000 because I mimicked the way transaction points were originally calculated in harvest. Yes, capped_amount is a better name, I will change it

src/seeds.history.cpp
IanMendozaJaimes: Yes, I you said below, it erases the smallest so it keeps the 2 greatest qualifying transactions. I did it that way because I didn't want to determine that in the `trxentry` action due to the reliability concerns, that's why it adds the transaction to the table and then in `savepoints` checks if it should be kept.

src/seeds.history.cpp
IanMendozaJaimes: Yes, that's correct. But it also could be negative values, those are calculated in order to adjust the total points, so even if they become negative, the total transaction points calculated will never be negative. I realized that a transaction could have a grater volume, but not necessarily give more transaction points, I thought we had to keep the 2 transactions with the biggest volume, is that right?

src/seeds.history.cpp
n13: Right so points are assigned on Seeds traded by the amounts are x10,000 ... makes sense.

src/seeds.history.cpp
IanMendozaJaimes: Mmm nope, it adjusts the total transaction points for "from", so  from_points could be negative, the total transaction points won't be. Here the algorithm is preferring the transaction that moves more Seeds rather than the transaction that gives more points. Would you like me to change that?

src/seeds.history.cpp
n13: Yes on the 2tx

I'll trust that it works on the others, not sure how they can be negative ;)

test/harvest.test.js
IanMendozaJaimes: Yes, that's right, my mistake sorry

test/harvest.test.js
IanMendozaJaimes: Sorry, I think I don't understand, do you mean this part: `'0'+memoprefix`?

n13:
Updated conflicts

@IanMendozaJaimes 

This action is used but was removed from history

ACTION organization::scoreorgs(name next) {
    require_auth(get_self());

    auto itr = next == ""_n ? organizations.begin() : organizations.lower_bound(next.value);
    if(itr != organizations.end()) {
        action(
            permission_level{contracts::history, "active"_n},
            contracts::history, "orgtxpoints"_n,
            std::make_tuple(itr -> org_name)
        ).send();
        itr++;
        if (itr != organizations.end()) {
            action next_execution(
                permission_level{get_self(), "active"_n},
                get_self(),
                "scoreorgs"_n,
                std::make_tuple(itr->org_name)
            );
            transaction tx;
            tx.delay_sec = 1; // change this to 1 to test
            tx.send(next.value+2, _self);
        }
    }
}

n13 (changes requested):
Need to fix call to "orgtxpoints" in history - call equivalent method?

IanMendozaJaimes:
Hello, I added all the changes requested. 
In the case of the action `scoreorgs`, I changed the action it calls, but `scoreorgs` is no longer needed, as all the transaction points calculations are done in harvest in the action `calctrxpts`. I forgot to remove the `scoreorgs` action sorry. In this commit I left it in case it is needed for other reason, but I removed it from scheduler's reset action.

n13 (approved):

PR #177, opened 2020-11-30
bugfix: don't check quorum for evaluate stage proposals

n13:
the quorum is dynamic and always changing depending on the number of proposals, number of eligible voters, and their vote power. Therefore we only check quorum on the initial vote. Once a proposal has been voted in and is in eval stage, we only check unity, in case someone changed their mind. But we don't check quorum.

@IanMendozaJaimes  please check

### DIFF
--- a/src/seeds.proposals.cpp
+++ b/src/seeds.proposals.cpp
@@ -231,11 +231,9 @@ void proposals::onperiod() {
 
         bool valid_quorum = false;
 
-        if (pitr->status != status_evaluate) {
-          // in evaluate status, we only check unity. 
+        if (pitr->status == status_evaluate) { // in evaluate status, we only check unity. 
           valid_quorum = true;
-        } else {
-          // in open status, quorum is calculated
+        } else { // in open status, quorum is calculated
           valid_quorum = utils::is_valid_quorum(voters_number, quorum, total_eligible_voters);
         }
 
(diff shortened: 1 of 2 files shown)


IanMendozaJaimes (approved):
It looks great!

src/seeds.proposals.cpp
```diff
@@ -225,11 +225,20 @@ void proposals::onperiod() {
...
         double fav = double(pitr->favour);
         bool passed = pitr->favour > 0 && fav >= double(pitr->favour + pitr->against) * majority;
-        bool valid_quorum = utils::is_valid_quorum(voters_number, quorum, total_eligible_voters);
         name prop_type = get_type(pitr->fund);
         bool is_alliance_type = prop_type == alliance_type;
         bool is_campaign_type = prop_type == campaign_type;
 
+        bool valid_quorum = false;
+
+        if (pitr->status != status_evaluate) {
```
IanMendozaJaimes, line 234: Sorry, I got confused.

I think the if statement is doing the opposite you described, right now, the quorum is immediately valid when the proposal is not in evaluate state and when the proposal is in evaluate state it is checking if the quorum is valid. I think this `pitr->status != status_evaluate` should be `pitr->status == status_evaluate`

IanMendozaJaimes (commented):
Sorry, I got confused at first, I think the if statement is doing the opposite you described.

src/seeds.proposals.cpp
n13: Thanks, you're right - that's why I need reviews ;)

n13:
Applied Ian's suggestions - deploying

PR #178, opened 2020-12-02
Calculate hourly mint quantity #140

IanMendozaJaimes:
- Added mint rate calculation in harvest contract
- Added tests

### DIFF
--- a/src/seeds.harvest.cpp
+++ b/src/seeds.harvest.cpp
@@ -927,14 +927,16 @@ void harvest::calcmintrate () {
   auto previous_day_temp = eosio::time_point_sec((day - (3 * utils::moon_cycle)) / 86400 * 86400);
   uint64_t previous_day = previous_day_temp.utc_seconds;
 
-  auto current_qev_itr = monthlyqevs.get(day, "no monthly qev found");
-  auto previous_qev_itr = monthlyqevs.get(previous_day, "no previous monthly qev found");
+  auto current_qev_itr = monthlyqevs.find(day);
+  auto previous_qev_itr = monthlyqevs.find(previous_day);
 
-  double volume_growth = double(current_qev_itr.qualifying_volume - previous_qev_itr.qualifying_volume) / previous_qev_itr.qualifying_volume;
+  if (current_qev_itr == monthlyqevs.end() || previous_qev_itr == monthlyqevs.end()) { return; }
 
-  int64_t target_supply = (1.0 + volume_growth) * previous_qev_itr.circulating_supply;
+  double volume_growth = double(current_qev_itr -> qualifying_volume - previous_qev_itr -> qualifying_volume) / previous_qev_itr -> qualifying_volume;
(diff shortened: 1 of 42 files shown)


src/seeds.harvest.cpp
```diff
@@ -905,3 +864,95 @@ ACTION harvest::setorgtxpt(name organization, uint64_t tx_points) {
...
+
+void harvest::calcmintrate () {
+  require_auth(get_self());
+
+  uint64_t day = utils::get_beginning_of_day_in_seconds();
+  auto previous_day_temp = eosio::time_point_sec((day - (3 * utils::moon_cycle)) / 86400 * 86400);
+  uint64_t previous_day = previous_day_temp.utc_seconds;
+
+  auto current_qev_itr = monthlyqevs.get(day, "no monthly qev found");
+  auto previous_qev_itr = monthlyqevs.get(previous_day, "no previous monthly qev found");
```
n13, line 931: When we are starting up we may not have a 3 quarter lookback. I guess it's ok for the action to do nothing in that case, but wouldn't we stop the scheduler with an error?

src/seeds.harvest.cpp
```diff
@@ -905,3 +864,95 @@ ACTION harvest::setorgtxpt(name organization, uint64_t tx_points) {
...
+
+  double volume_growth = double(current_qev_itr.qualifying_volume - previous_qev_itr.qualifying_volume) / previous_qev_itr.qualifying_volume;
+
+  int64_t target_supply = (1.0 + volume_growth) * previous_qev_itr.circulating_supply;
+
+  int64_t delta = target_supply - current_qev_itr.circulating_supply;
+
```
n13, line 938: delta could be negative (shrinking) - have to handle here, can just do nothing for the time being

n13 (changes requested):
added two comments, looks good otherwise

src/seeds.harvest.cpp
IanMendozaJaimes: Yes, I will replace it with a .find

src/seeds.harvest.cpp
IanMendozaJaimes: Ok, do you want me to set delta to zero if it becomes negative? I left it like that just to store the information, because in the distribute harvest ticket, the harvest is not calculated if delta is less than zero

IanMendozaJaimes:
Now if there is no enough information to calculate the mint rate the `calcmintrate` action just exits.
Now if the mint rate becomes less than 0 it is set to 0.

src/seeds.harvest.cpp
n13: Oh... damn... I guess your original implementation was better then

n13: I reverted that again so mint rate can be negative. 

Your solution was better.

n13 (approved):

Issue #141, opened 2020-10-21
Distribute test harvest


n13:
Distribute test harvest every hour = this should get invoked from #140 once we know the rate...

Distritbute into the correct buckets with the correct amounts

juliolrmonteiro:
Distribute another TEST SEEDS token instead, not actual SEEDS yet, token contract using "test.seeds"

IanMendozaJaimes:
Hello, I have some questions:
- In the documentation says that one part of the harvest belongs to Global G-DHO Harvest, can you please tell me who is this account?
- Just to make sure I understood: 
  Let's say the `mint rate` is 1000 Seeds/hour and the distributions are like in the constitution: Residents/Citizens(30%), Bioregions(30%), Orgs(20%), Global(20%).
  
  Then 1000 Seeds have to be issued per hour, and each bucket will receive hourly:
  - Residents/Citizens => 300 Seeds
  - Bioregions => 300 Seeds
  - Orgs => 200 Seeds
  - Global => 200 Seeds
  
  Now, let's suppose there are 150 residents/citizens in total, but only 2 have a rank 99, so first I need to determine how much Seeds each rank deserves, using:
  99x + 98x + 97x + ... + 1x = 150
  In this example x = 0.0303, so the rank 99 deserves 99 * 0.0303 = 2.9997 Seeds. As there are 2 users in the rank 99, then each user receives 2.9997 / 2 = 1.4998 Seeds
  
  And the same logic applies for orgs and bioregions
  
  Is all the above right?

n13:
On 1 - G-DHO is the harvest contract... this pool will be used to find proposals.

IanMendozaJaimes:
Hello @n13, I have some other questions:

- Is contribution score for orgs been calculated in the action `calccss`?, If so, I think it is missing the transaction score for orgs, is it ok to change it in this ticket?

- I have noticed that the contribution score for bioregions is not being calculated, for the moment, what should I do with the part of the harvest belonging to bioregions?

> Seeds each rank deserves, using:
99x + 98x + 97x + ... + 1x = 300 Seeds
In this example x = 0.0606, so the rank 99 deserves 99 * 0.0606 = 5.9994 Seeds. As there are 2 users in the rank 99, then each user receives 5.9994 / 2 = 2.9997 Seeds
[...]

PR #180, opened 2020-12-04
Distribute test harvest #141

IanMendozaJaimes:
- Added the creation of the test token TESTS and the global account gdho.seeds for local net only.
- Modified `rankcss` to perform the sum of all ranks in this action
- Added harvest distribution actions
- Added permission minttst to harvest contract so it can use the `minttest` action
- Added tests

### DIFF
--- a/src/seeds.harvest.cpp
+++ b/src/seeds.harvest.cpp
@@ -973,16 +973,20 @@ void harvest::calcmintrate () {
   auto previous_day_temp = eosio::time_point_sec((day - (3 * utils::moon_cycle)) / 86400 * 86400);
   uint64_t previous_day = previous_day_temp.utc_seconds;
 
-  auto current_qev_itr = monthlyqevs.get(day, "no monthly qev found");
-  auto previous_qev_itr = monthlyqevs.get(previous_day, "no previous monthly qev found");
+  auto current_qev_itr = monthlyqevs.find(day);
+  auto previous_qev_itr = monthlyqevs.find(previous_day);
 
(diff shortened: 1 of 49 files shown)


src/seeds.harvest.cpp
```diff
@@ -905,3 +910,278 @@ ACTION harvest::setorgtxpt(name organization, uint64_t tx_points) {
...
+  print("mint rate:", quantity, "\n");
+
+  token::issue_action_test t_issue{contracts::token, { contracts::token, "minttst"_n }};
+  t_issue.send(get_self(), quantity, memo);
+
+  double users_percentage = config_get("hrvst.users"_n) / 1000000.0;
+  double bios_percentage = config_get("hrvst.bios"_n) / 1000000.0;
+  double orgs_percentage = config_get("hrvst.orgs"_n) / 1000000.0;
+  double global_percentage = config_get("hrvst.global"_n) / 1000000.0;
+
```
n13, line 1057: I think we need to add a "escrow" percentage also - some of the Seeds are supposed to get paid out with escrow but I am not sure yet how that really works. Since escrow is based on people... not everyone gets the same...  IDK

No action item, just thinking out loud here. We can release this as is here.

n13 (changes requested):
minor permissions set up changes requested

IanMendozaJaimes:
I changed the permissions as you suggested.

n13:
> I changed the permissions as you suggested.

Ok I removed the key permission also, not needed anymore. All good.

n13 (approved):

Issue #58, opened 2020-05-12
Basic Delegate Trust


n13:
## Spec

Delegate Trust
Citizens may delegate voting rights to another Citizen. 
This earns Citizens rewards as described in the Reputation section of this paper. 
Delegations mimic the relative voting of the delegatee. 
Example: If delegatee casts 1/10 of their Trust Tokens on a proposal it follows with 1/10 of the Trust Tokens of the delegator.
 [%] of Trust Tokens cast delegatee = [%] of Trust Tokens cast of delegator 

## Implementation

Could do this relatively easily with a delegations table on the proposals contract

If one has made delegations, they can't vote themselves
delegations can have a weight 1 - 100 (fixed to 100% for the first implementation)

juliolrmonteiro:
Delegate votes need to be by Categories (Alliances, Proposals, Referendums for now)
You can only assign one Citizen to vote for you on that category.
You voting power is fully available on all Categories (not split)

IanMendozaJaimes:
Hello I have some questions:

- How long does voice delegation last? I mean, a user needs to call another action to cancel voice delegation or does it have a duration period?

- Is there any limit of delegators a user can have? I mean, the user A can be the delegatee of any number of users?

n13:
Yes there needs to be a cancel function 

The user can either assign another trust delegate or cancel delegation any time 

It lasts forever although I think the trust delegate should also be able to cancel - if they don’t want to do it anymore. 

No limit on number of users

PR #185, opened 2020-12-08
Basic Delegate Trust #58

IanMendozaJaimes:
- Added functions to delegate and cancel delegation of trust
- Added tests
- Added new parameter `votedel.mul` to settings

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -89,11 +89,11 @@ CONTRACT proposals : public contract {
 
       ACTION initnumprop();
 
-      ACTION delgatetrust(name delegator, name delegatee, name scope);
+      ACTION delegate(name delegator, name delegatee, name scope);
 
-      ACTION mimicvote(name delegatee, name scope, uint64_t proposal_id, double percentage_used, name option, uint64_t chunksize);
+      ACTION mimicvote(name delegatee, name delegator, name scope, uint64_t proposal_id, double percentage_used, name option, uint64_t chunksize);
 
-      ACTION cncldeltrst(name delegator, name scope);
+      ACTION undelegate(name delegator, name scope);
 
   private:
       symbol seeds_symbol = symbol("SEEDS", 4);
@@ -141,7 +141,7 @@ CONTRACT proposals : public contract {
       void send_to_escrow(name fromfund, name recipient, asset quantity, string memo);
       void burn(asset quantity);
       void update_voice_table();
-      void vote_aux(name voter, uint64_t id, uint64_t amount, name option, bool is_new, bool is_mimic);
+      void vote_aux(name voter, uint64_t id, uint64_t amount, name option, bool is_new, bool is_delegated);
       bool revert_vote (name voter, uint64_t id);
       void change_rep(name beneficiary, bool passed);
       uint64_t get_size(name id);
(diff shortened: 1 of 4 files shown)


include/seeds.proposals.hpp
```diff
@@ -89,6 +89,12 @@ CONTRACT proposals : public contract {
 
       ACTION initnumprop();
 
+      ACTION delgatetrust(name delegator, name delegatee, name scope);
+
+      ACTION mimicvote(name delegatee, name scope, uint64_t proposal_id, double percentage_used, name option, uint64_t chunksize);
+
+      ACTION cncldeltrst(name delegator, name scope);
```
n13, line 96: delegate()
undelegate()

for these functions?

include/seeds.proposals.hpp
```diff
@@ -242,6 +251,16 @@ CONTRACT proposals : public contract {
 
         uint64_t primary_key()const { return account.value; }
       };
+
+      TABLE delegate_trust_table { // scoped by proposal's category (alliance, campaign, etc)
+        name delegator;
+        name delegatee;
+        double weight;
```
n13, line 258: cool!

src/seeds.proposals.cpp
```diff
@@ -795,8 +804,7 @@ bool proposals::revert_vote (name voter, uint64_t id) {
   return false;
 }
 
-void proposals::vote_aux (name voter, uint64_t id, uint64_t amount, name option, bool is_new) {
-  require_auth(voter);
+void proposals::vote_aux (name voter, uint64_t id, uint64_t amount, name option, bool is_new, bool is_mimic) {
```
n13, line 807: why "mimic" and not "delegate" ?

src/seeds.proposals.cpp
```diff
@@ -1271,3 +1295,131 @@ ACTION proposals::initnumprop() {
...
+    ditr++;
+    count++;
+  }
+
+  if (ditr != deltrusts_by_delegatee.end() && ditr -> delegatee == delegatee) {
+    action next_execution(
+      permission_level{get_self(), "active"_n},
+      get_self(),
+      "mimicvote"_n,
+      std::make_tuple(delegatee, scope, proposal_id, percentage_used, option, chunksize)
```
n13, line 1399: Will this not loop forever when size > chunk size? 

The second call signature is the exact same as the first, no pointers are moving. That's a lot of votes..

n13: Suggest to add a test with 2 delegations and chunk size 1

n13 (changes requested):

include/seeds.proposals.hpp
IanMendozaJaimes: I changed the names of the functions

include/seeds.proposals.hpp
IanMendozaJaimes: jaja thanks, but I am not using the weight yet, I wasn't sure where I had to use it.

src/seeds.proposals.cpp
IanMendozaJaimes: I changed the name of the variable

src/seeds.proposals.cpp
IanMendozaJaimes: You are completely right, what a horrible mistake I made here. I fixed in the commit

include/seeds.proposals.hpp
n13: It's cool weight is in there

n13 (approved):

PR #186, opened 2020-12-08
Onboard new guardians into Seeds

7flash:
The idea is that instead of having to use existing guardians, people can sign up new users to be guardians.

### DIFF
--- a/include/seeds.guardians.hpp
+++ b/include/seeds.guardians.hpp
@@ -1,5 +1,6 @@
 #include <eosio/eosio.hpp>
 #include <eosio/system.hpp>
+#include <eosio/crypto.hpp>
 #include <tables/user_table.hpp>
 #include <abieos_numeric.hpp>
 #include <contracts.hpp>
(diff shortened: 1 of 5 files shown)


include/seeds.guardians.hpp
```diff
@@ -22,6 +28,18 @@ CONTRACT guardians : public contract
 
     ACTION reset();
 
+    ACTION init(name protectable_account, checksum256 invite_hash);
+
+    ACTION join(name guardian_account, checksum256 invite_secret);
```
n13, line 33: Not sure what the secret is for, or how having a secret would improve things?

- A secret can only be used once - after that it's on chain and no longer a secret
- A secret on the side of the original account would have to be stored somewhere. If they lose their private key they can also lose their secret. If the two are interchangeable because a secret can restore a private key then there's no benefit to having both. 
- If the protected account lost access to their wallet, they will have lost access to the secret as well.
- Same for the guardian accounts - either they have access to their account and secret, or they don't have access to either.

n13 (changes requested):

include/seeds.guardians.hpp
n13: Nevermind... secret is to onboard new guardian accts

test/guardians.test.js
```diff
@@ -49,11 +50,27 @@ const addActorPermission = async (target, targetRole, actor, actorRole) => {
   }
 }
 
-function sleep(ms) {
-  return new Promise(resolve => setTimeout(resolve, ms));
-}
+describe('guardians', async a => {
```
n13, line 53: magic.....

n13 (commented):
We need to talk about this... I dont like it's a lot more code and I don't like changing code that was much simpler because I could understand it all and see it didn't have any holes. Now... making this water tight is super important, for obvious reasons.

n13:
closing for now... I think we found a way to do this without any smart contract changes

Issue #160, opened 2020-11-12
Separate org CS score from normal accounts CS score


n13:
Harvest contract a contribution score (CS) table

Currently this table contains CS for organizations as well as accounts

Make it so it only contains CS for individuals

Organizations need their own CS table - could be scope

n13:
We might need to separate out all org tables from individual tables

See harvest contract here on this method
void harvest::ranktx(uint64_t start_val, uint64_t chunk, uint64_t chunksize, name table) {

juliolrmonteiro:
This seems to be done already, CS seems to be calculate at:
```C++
void harvest::calccs(uint64_t start_val, uint64_t chunk, uint64_t chunksize) {
  require_auth(_self);

  check(chunksize > 0, "chunk size must be > 0");

  uint64_t total = utils::get_users_size();
  auto uitr = start_val == 0 ? users.begin() : users.lower_bound(start_val);
  uint64_t count = 0;

  while (uitr != users.end() && count < chunksize) {
    calc_contribution_score(uitr->account);
    count++;
    uitr++;
  }
```

So it seems it only iterates over users, not over orgs.

juliolrmonteiro:
This seems to be already done:

```C++
ACTION harvest::setorgtxpt(name organization, uint64_t tx_points) {
  require_auth(get_self());

  tx_points_tables orgtxpoints(get_self(), "org"_n.value);
```
The table is scoped differently here.

n13:
Done

Issue #151, opened 2020-11-09
Change CBP rewards for referrals to vouching!


n13:
## Spec

VOUCHING - gets community building points

Referrals - gets reputation points

This is currently wrong where CBP is only given for referrals

Change the names of these settings also

  // community buiding points for referrer when user becomes resident
  confwithdesc(name("vou.cbp1.ind"), 1, "Community buiding points for referrer when user becomes resident", high_impact);

  // community buiding points for referrer when user becomes citizen
  confwithdesc(name("vou.cbp2.ind"), 1, "Community buiding points for referrer when user becomes citizen", high_impact);

## Current State
- irg gets cbp when org is a referrer - orgs don't vouch so this makes sense
- org does not get rep points for this - that's pribably also OK - do orgs even get normal rep points, or only regen points as voted by citizens?
- individuals get rep points and cbp - this should change so all vouchers get cbp - we already iterate over vouch in vouchreward, so we can just add that there. 
- currently orgs also get entries in the cbp table in accounts - they should not be in there at all. 
 

## Migration
Assign CBP points for vouches for residents and citizen

Give these CBP to all vouchers, except for the referrer

Reasoning: A referrer is always also a voucher - meaning who referred me, also vouched for me

The referrer already received their CBP points reward in the old method, so they are OK - this is correct

Only the other vouchers did not receive any, so for those we play catch up

For a citizen, they get resident reward and citizen reward. For resident, they only get resident reward.

juliolrmonteiro:
Doing the Migration for historical vouching is a bonus :)

juliolrmonteiro:
Hey @n13, I was reviewing the constitution at https://docs.google.com/document/d/1C4w9Ol8VGabCIcQDVPDrwcTRoJXBqhrb7VjslwQbUGU/edit#

And from what I understood from the text you only get CPB when someone you invited (referral) becomes resident or citizen, not when you VOUCH for them (page 46), did this change somewhere?

From what I get at the Constituion, the current code is correct as it issues CBP only on `accounts::refreward` when the referred user becomes resident or citizen.

Can you please doublecheck if this change is needed?

n13:
Yeah this issue was about that all who have vouched should receive this reward

Not just the referrer as it’s currently implemented

Issue #150, opened 2020-11-09
Const - add vis.plant setting as min planted for invitations


n13:
This is currently hard-coded - at 5.

Spec: Make it a setting.

juliolrmonteiro:
Should it be named "vis.plant" or "fee.invitation", because it's not used only by visitors, right?

juliolrmonteiro:
What about "invitation.planted"

juliolrmonteiro:
Check if `sow_quanted` is above the minimum in `_invite`.
Make sure people ar not allowed to unplant this min quantity?

PR #189, opened 2020-12-10
Small fixes for alignment with constitution

juliolrmonteiro:
The main issue that was fixed here is #151, with tests, but no migration yet.

Also added the needed settings for #150 (but besides that, the change to actually use this setting is at Passport and Light Wallet)

Checked issue #160 and it's already separated, from what I can reckon.

### DIFF
--- a/src/seeds.accounts.cpp
+++ b/src/seeds.accounts.cpp
@@ -98,7 +98,6 @@ void accounts::vouch(name sponsor, name account) {
   name account_status = uitra->status;
 
   check(sponsor_status == name("citizen") || sponsor_status == name("resident"), "sponsor must be a citizen or resident to vouch.");
-
   _vouch(sponsor, account);
 }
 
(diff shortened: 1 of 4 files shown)


src/seeds.accounts.cpp
```diff
@@ -259,25 +263,12 @@ void accounts::refreward(name account, name new_status) {
...
-    });
-  } else {
-    cbs.emplace(_self, [&](auto& item) {
-      item.account = referrer;
-      item.community_building_score = community_building_points;
-      item.rank = 0;
-    });
-    size_change("cbs.sz"_n, 1);
-  }
+  add_cbs(referrer, community_building_points);
```
n13, line 271: The referrer should not get additional CBP - they are already included with the vouch rewards?

src/seeds.settings.cpp
```diff
@@ -102,10 +106,10 @@ void settings::reset() {
   // =====================================
 
   // community buiding points for referrer when user becomes resident
-  confwithdesc(name("refcbp1.ind"), 1, "Community buiding points for referrer when user becomes resident", high_impact);
+  confwithdesc(name("vou.cbp1.ind"), 1, "Community buiding points for voucher when user becomes resident", high_impact);
 
   // community buiding points for referrer when user becomes citizen
-  confwithdesc(name("refcbp2.ind"), 1, "Community buiding points for referrer when user becomes citizen", high_impact);
```
n13, line 108: Some code still refers to the old value here so IDK how this is passing unit tests.

      const name cbp_reward_citizen = "refcbp2.ind"_n; // hpp file

      name cbp_param = is_citizen ? cbp_reward_citizen : cbp_reward_resident; // cpp file

I made another comment down there, let's delete the code that hands out CBP for referrals (they get them with the vouching already). Then delete that variable which refers to a non-existent setting.

n13 (changes requested):
See comments in code, I think there's bugs here.

Let me know if I'm wrong ;)

src/seeds.accounts.cpp
juliolrmonteiro: Got it.

src/seeds.settings.cpp
juliolrmonteiro: Done. Check that the accounts unit test called "test remove" was checking if the used had CBP after the seconduser was removed, but since they were not `invited` (and thus vouched), they did not get CPB.

juliolrmonteiro:
passed: 47,  failed: 0  of 47 tests  (56.9s)

n13 (approved):

PR #194, opened 2020-12-16
Bugfix/actives refactoring

n13:
* Refactor actives list for the fact we don't need the "active" flag since the time stamp contains the same information

This is possible now because users can't jump in and out of being citizens - no more demoting citizens or coming back.

So we really only need the information as to when someone voted last to determined if we should consider them active.


IanMendozaJaimes (approved):
It looks great!, it is cleaner now.

PR #195, opened 2020-12-16
Feature/deploy scheduler

n13:
Added 2 commands to seeds.js

updateSettings: deploys and resets the settings contract
updateScheduler: deploys and updates the scheduler contract (non-destructive)


n13:
Just some deployment utilities

Issue #153, opened 2020-11-09
Transaction Point Cycle Length setting


n13:
Have a floating point setting to consider transactions from X cycles back for history for Transactions Points for individuals and orgs.

We look back 3 cycles to calculate transaction scores

'3' should be a setting that could be changed, example to 2.5

Needs new floating point settings

PR #196, opened 2020-12-17
Transaction Point Cycle Length setting #153

IanMendozaJaimes:
Added setting to control the number of cycles to take into account for transaction points calculation


n13 (approved):
Looking good

Issue #162, opened 2020-11-12
Time limit on changing bioregions


n13:
1 cycle delay

There is a [1 full cycle] delay when changing a bioregion. If a member changes bioregions during a voting cycle they will need to wait until that cycle is finished and the next full cycle has passed before they can vote on proposals.

IanMendozaJaimes:
Hello, I have some questions:
1. If a user has  a role, does the user need to call first leaverole action? or  does the user leave the role automatically?
2. Does the delay also applies when changing bioregions? I mean, if the user A changes its bioregion, then the user A has to wait until the end of the next period to be able to change its bioregion again
3. I suppose that if the user A has delegated its trust, then even if A's delegatee votes the user A's delegated vote should not happen

n13:
> Hello, I have some questions:
> 
> 1. If a user has  a role, does the user need to call first leaverole action? or  does the user leave the role automatically? I think this could be applied more generally in the action `leave`, currently this action removes the user from the bioregion members table, but it leaves the user in the role table so even if the user is not a member it still has a role in the bioregion which might be a bug.

Certainly a bug.

Maybe a user who has a role cannot leave the BR until they first leave their role. To make sure the contract data is consistent. 


> 2. Does the delay also applies when changing bioregions? I mean, if the user A changes its bioregion, then the user A has to wait until the end of the next period to be able to change its bioregion again.

Yes definitely. Any time they change, there's that delay.

> 3. Currently it seems a user doesn't have a default bioregion, when a user joins for the first time to a bioregion (i.e. the first time a user enters in the members table), does the user has to wait the same period to be able to vote? or in this case as it's the first time the user can vote immediately?

No. Only on changing. No wait on first bioregion.

It seemed pretty clear in our minds - the purpose of this is to stop people from switching bioregions when convenient, example to hop in and out of a vote.

But we still want to make it possible to change bioregion. 

So there's a delay of 1 cycle to switch

What's the best mechanism to implement is what we're trying to find out here.

> 4. For this variable: [1 full cycle], I suppose it has to be flexible, for example it could be set to [1.5 full cycle], in the case it is set to 1.5, if the user A changed bioregions in the middle of cycle, then user A needs to wait for that cycle to end, and wait 1.5 cycles to be able to vote, so user A can vote until the second half of the third cycle, right?
> 5. I suppose that if the user A has delegated its trust, then even if A's delegatee votes the user A's delegated vote should not happen

[...]

PR #199, opened 2020-12-21
Time limit on changing bioregions #162

IanMendozaJaimes:
- Added delay when changing bioregions
- Fixed bug that allows a user to leave the bioregion without removing the role

I added the table `delay_table` to determine whether is the first time the user joins to some bioregion and the delay the user should wait in case it's not the first time. The `members` table is not enough for two reasons: 1) it is not possible to determine it is the first time the user joins only with the timestamp, so a migration would be needed to add the extra parameter. 2) This is the strongest, the methods `leave` and `removemember` delete the member from the `members` table, so these methods should be modified to keep the member in the table, but I think that wasn't needed and it was best just to have another table.

Another thing, I added the delay when changing bioregions in the method `join`, so a user can't switch opportunistically, but I didn't add the delay when voting because so far the proposals don't belong to any bioregion, they are global, and I don't have a clear idea about how bioregional proposals will be, however with the `delay_table` will be easy to add the voting delay whenever it is needed.


src/seeds.bioregion.cpp
```diff
@@ -160,15 +165,33 @@ ACTION bioregion::join(name bioregion, name account) {
...
+
+    auto ditr = biodelays.find(account.value);
+    if (ditr == biodelays.end()) {
+        biodelays.emplace(_self, [&](auto & item){
             item.account = account;
+            item.apply_vote_delay = false;
+            item.joined_date_timestamp = now;
+        });
+    } else {
+        check(ditr -> joined_date_timestamp < now - (utils::moon_cycle * config_float_get("bio.vote.del"_n)), "user needs to wait until the delay ends");
```
juliolrmonteiro, line 182: Since only 1 float config is being used, would it be just more clear to add a variable to store the config, instead of a function?

juliolrmonteiro (approved):
Looks good to me, just a suggestion for a small optional change.

src/seeds.bioregion.cpp
```diff
@@ -160,15 +165,33 @@ ACTION bioregion::join(name bioregion, name account) {
     auto mitr = members.find(account.value);
 
     if (mitr != members.end()) {
-        check(false, "user exists TODO allow users to switch");
-    } else {
-        members.emplace(_self, [&](auto & item) {
-            item.bioregion = bioregion;
+        if (mitr -> bioregion == bioregion) { return; }
+        leave(mitr -> bioregion, account);
```
n13, line 169: Join should still fail - the user needs to explicitly leave the bioregion

It doesn't make sense to "leave" automatically, but then fail, because they can't yet join due to the delay?!

So what I mean to say is - this join will fail in this case as leaving and immediately joining won't work, so might as well just fail here, instead of "leave" - same result.

src/seeds.bioregion.cpp
n13: @juliolrmonteiro config_get is our standard, as it takes care of the case where the setting was not yet set.

src/seeds.bioregion.cpp
```diff
@@ -231,6 +254,10 @@ void bioregion::remove_member(name account) {
     
     check(mitr != members.end(), "member not found");
 
+    roles_tables roles(get_self(), mitr -> bioregion.value);
+    auto ritr = roles.find(account.value);
+    check(ritr == roles.end(), "user can not leave bioregion because it has a role in it");
+
```
n13, line 260: Could either remove the role, or have them remove the role first - I am fine with this, good catch

Clients can automate this , e.g. they can check if there's roles, and add a leave the role action before the leave action

n13: Actually if a member is removed by an admin, they should also be removed from the role?! Hmm.

src/seeds.settings.cpp
```diff
@@ -290,7 +294,7 @@ void settings::confwithdesc(name param, uint64_t value, string description, name
   auto citr = config.find(param.value);
 
   auto fitr = configfloat.find(param.value);
-  check(fitr == configfloat.end(), "this parameter is defined as floating point");
+  check(fitr == configfloat.end(), param.to_string() + ", this parameter is defined as floating point");
```
n13, line 297: great! ;)

n13 (changes requested):
I think the code in join needs to change were an existing BR member is calling leave() - can just fail

Unless I am missing something

src/seeds.bioregion.cpp
IanMendozaJaimes: Yes, you are right. I fixed that

src/seeds.bioregion.cpp
IanMendozaJaimes: Ok, it is better to have the role removed automatically

IanMendozaJaimes:
I added the requested changes

n13 (approved):
Great

PR #200, opened 2020-12-21
Bug fix in vote delegation

IanMendozaJaimes:
The way the action `mimicvote` was calling the function `vote_aux` was in the same transaction so if some delegator for some reason can not vote, then the whole action will fail. 

One example of that could be when the user A voted trust for a proposal that is now in evaluation phase, then user B delegates its voice to user A, and then user A votes distrust. This will cause a fail because B can't vote for the proposal.
 
That's why I added the action `voteonbehalf`, so it could be called in a deferred transaction inside `mimicvote`, in that way if a particular vote fails the rest of the mimicked votes don't.

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -54,6 +54,8 @@ CONTRACT proposals : public contract {
 
       ACTION neutral(name user, uint64_t id);
 
+      ACTION voteonbehalf(name voter, uint64_t id, uint64_t amount, name option);
+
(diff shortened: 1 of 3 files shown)


juliolrmonteiro (approved):
Looks good to me.

IanMendozaJaimes:
I forgot to mention that, but actually the example I gave is in the test cases, the fifthuser's vote is failing for that reason but the rest of delegators are fine

PR #201, opened 2020-12-22
handle HUSD

n13:
handle HUSD same way we used to handle TLOS

differences: HUSD is 2 digits of precision and comes from husd.hypha token contract.


7flash (approved):
Thanks! :money_with_wings:

PR #202, opened 2020-12-23
Bracelet management contract (bag of seeds) #197

IanMendozaJaimes:
- Added new contract bracelet
- Added tests

The `transfer` action, enforces the usage of the `bracelet` permission. The `bracelet` permission should be created by the user as well as the linkAuth. There is an example of this in the tests.

I didn't create the contract in helper.js for testnet nor mainnet, just in case

### DIFF
--- a/include/seeds.bracelet.hpp
+++ b/include/seeds.bracelet.hpp
@@ -53,7 +53,7 @@ CONTRACT bracelet : public contract {
     TABLE balance_table {
       name account;
       asset balance;
-      bool is_freezed;
+      bool is_frozen;
 
       uint64_t primary_key() const { return account.value; }
     };
--- a/include/seeds.bracelet.hpp
+++ b/include/seeds.bracelet.hpp
@@ -1,0 +1,80 @@
+#include <eosio/eosio.hpp>
+#include <eosio/asset.hpp>
+#include <eosio/transaction.hpp>
+#include <eosio/singleton.hpp>
+#include <seeds.token.hpp>
+#include <contracts.hpp>
+#include <utils.hpp>
+#include <tables/user_table.hpp>
+#include <tables/config_table.hpp>
+
+using namespace eosio;
+using namespace utils;
+using std::string;
+
+CONTRACT bracelet : public contract {
+
+  public:
+
+    using contract::contract;
+    bracelet(name receiver, name code, datastream<const char*> ds)
+      : contract(receiver, code, ds),
+        balances(receiver, receiver.value),
+        users(contracts::accounts, contracts::accounts.value)
+        {}
+
+    ACTION reset();
+
+    ACTION deposit(name from, name to, asset quantity, string memo);
+
+    ACTION freeze(name account);
+
+    ACTION unfreeze(name account);
+
+    ACTION withdraw (name account, asset quantity);
+
+    ACTION transfer(name from, name to, asset quantity, string memo);
+
+
+  private:
+
+    void check_user(name account);
+    void _deposit(asset quantity);
+    void init_balance(name account);
+    void add_balance(name account, asset quantity);
+    void sub_balance (name account, asset quantity);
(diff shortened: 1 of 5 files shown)


src/seeds.bracelet.cpp
```diff
@@ -0,0 +1,141 @@
...
+  _transfer(to, quantity, memo);
+}
+
+void bracelet::init_balance (name account) {
+  auto bitr = balances.find(account.value);
+  if (bitr == balances.end()) {
+    balances.emplace(_self, [&](auto & item){
+      item.account = account;
+      item.balance = asset(0, utils::seeds_symbol);
+      item.is_freezed = false;
```
juliolrmonteiro, line 78: I would rename this to "is_frozen"

include/seeds.bracelet.hpp
```diff
@@ -0,0 +1,80 @@
...
+    void check_freeze(name account);
+
+    DEFINE_USER_TABLE
+
+    DEFINE_USER_TABLE_MULTI_INDEX
+
+    TABLE balance_table {
+      name account;
+      asset balance;
+      bool is_freezed;
```
juliolrmonteiro, line 56: Better rename to "is_frozen" :)

src/seeds.bracelet.cpp
```diff
@@ -0,0 +1,141 @@
...
+    item.balance -= quantity;
+  });
+
+  balances.modify(bcitr, _self, [&](auto & item){
+    item.balance -= quantity;
+  });
+}
+
+void bracelet::_deposit (asset quantity) {
+  check_asset(quantity);
```
juliolrmonteiro, line 121: Should this be `utils::check_asset`?

src/seeds.bracelet.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  token::transfer_action action{contracts::token, {_self, "active"_n}};
+  action.send(_self, contracts::bank, quantity, "");
```
juliolrmonteiro, line 124: Looks like the memo is empty here on this transfer, should it mention it's a "bracelet deposit"?

src/seeds.bracelet.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+void bracelet::_transfer (name beneficiary, asset quantity, string memo) {
+  utils::check_asset(quantity);
+  token::transfer_action action{contracts::token, {contracts::bank, "active"_n}};
+  action.send(contracts::bank, beneficiary, quantity, memo);
+}
+
+void bracelet::check_freeze (name account) {
+  auto bitr = balances.get(account.value, "bracelet: no balance object found");
+  check(!bitr.is_freezed, "bracelet: account is freezed");
```
juliolrmonteiro, line 140: Just rename all instances of "freezed" to "frozen" ;)

juliolrmonteiro (changes requested):
This is very simple and well written, just small checks and requesting var names review.

src/seeds.bracelet.cpp
IanMendozaJaimes: Yes, you're right, is_frozen is a better name

src/seeds.bracelet.cpp
IanMendozaJaimes: ok, I will change it. What about the memo in the `withdraw` action?, I left it empty as well for the moment

src/seeds.bracelet.cpp
IanMendozaJaimes: yes, I forgot the namespace, I'll change it

IanMendozaJaimes:
I made the requested changes.

src/seeds.bracelet.cpp
juliolrmonteiro: Perfect, that's good for now. :)

juliolrmonteiro (approved):
Looks good to me. Thanks!

n13 (approved):

PR #203, opened 2020-12-28
Gratitude token contract

juliolrmonteiro:
This is the first implementation of the SEEDS gratitude token.

The basic idea is that you can allocate some funds to the `gratz.seeds` contract, then everyone gets some gratitude tokens to share. At the end of the round, the amount of SEEDS resting at the contract gets split among everyone that received gratitude during the round, in proportion to how much they got, for example:

If the round volume was 1000 gratitude tokens, and you got 100 in total, you get 100/1000 (1/10) of the stash of SEEDS.

When the round ends, all gratitude tokens available for giving are reset to the original quantity, and all earned tokens are back to zero for all users.

Still need to schedule the "newround()" action on seeds.scheduler, for once every cycle.

- contracts to deploy: settings, gratitude
- permissions added Y `gratitude@active`
- new accounts created Y `gratz.seeds`
- settings changed Y
- scheduler changed N (not yet)
- migrate functions required (none yet)

### DIFF
--- a/src/seeds.gratitude.cpp
+++ b/src/seeds.gratitude.cpp
@@ -31,6 +31,11 @@ ACTION gratitude::reset () {
 ACTION gratitude::give (name from, name to, asset quantity, string memo) {
   require_auth(from);
 
+  check( from != to, "gratitude: cannot give to self" );
+  check( is_account( to ), "gratitude: to account does not exist");
+  check_asset(quantity);
+  check( quantity.amount > 0, "gratitude: must give positive quantity" );
+
   // Should create balances if not there yet
   init_balances(to);
   init_balances(from);
@@ -38,7 +43,7 @@ ACTION gratitude::give (name from, name to, asset quantity, string memo) {
   sub_gratitude(from, quantity);
   add_gratitude(to, quantity);
 
-  increase_volume(quantity.amount);
+  update_stats(from, to, quantity);
 }
 
 ACTION gratitude::newround() {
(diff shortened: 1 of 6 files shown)


src/seeds.gratitude.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  auto contract_balance = eosio::token::get_balance(contracts::token, get_self(), seeds_symbol.code());
+  uint64_t tot_accounts = get_size("balances.sz"_n);
+  uint64_t volume = get_current_volume();
+
+
+  auto balances_by_received = balances.get_index<"byreceived"_n>();
+
+  auto bitr = balances_by_received.begin();
+  while (bitr != balances_by_received.end()) {
```
n13, line 61: This will only work up to a balances size of about 200 - 400, past that it won't work

Which may not be an issue- certainly won't run into this any time soon if we use this inside hypha.

src/seeds.gratitude.cpp
```diff
@@ -0,0 +1,197 @@
...
+    uint64_t my_received = bitr->received.amount;
+    balances_by_received.modify(bitr, _self, [&](auto& item) {
+        item.received = asset(0, gratitude_symbol);
```
n13, line 64: Wondering - if you iterate though "by received", then change the received value to 0 - is this safe to use with a c++ iterator?

src/seeds.gratitude.cpp
```diff
@@ -0,0 +1,197 @@
...
+  // Should always work because reset always creates first round
+  return stitr->volume.amount;
+}
+
+void gratitude::update_stats(name from, name to, asset quantity) {
+  auto stitr = stats.rbegin();
+  auto round_id = stitr->round_id;
+
+  auto stitr2 = stats.find(round_id);
+  auto oldvolume = stitr2->volume.amount;
```
n13, line 93: stitr2 is now == stitr - why not just use sitr?

I don't understand this code...

n13 (commented):
Added some comments, can't say I understand everything in this contract

Maybe add to scheduler.... we should deploy this so we can play around with it

src/seeds.gratitude.cpp
juliolrmonteiro: Well, in `newround` we don't need to iterate by received, anyway, since we're paying every account with higher than 0 balance.

src/seeds.gratitude.cpp
juliolrmonteiro: Fixed.

src/seeds.gratitude.cpp
juliolrmonteiro: This is to find the `round_id` of the last round (`stats.rbegin()`), so that's why I need 2 iterators.

n13 (approved):
I can't really verify this is all working, as it's too much code, and I am not 100% clear on the specs

But I verified it is not interfering so I feel we can just deploy this and play around with it?

Issue #207, opened 2021-01-11
Quorum is generated from the avg voice cast form the last 2 cycles


n13:
## Scope

Main change is to take the avg votes cast in the last X cycles where X = 2 as a basis for the quorum.

Example: 
4500 votes were cast in the last cycle
4200 votes the cycle before
If cycle < 0, take 50% of available votes as a basis (not happening anyway but maybe in tests and also if someone else deploys it)

Avg = 4350 == quorum base

Then we calculate the quorum %, example it's 20%
4350 * 0.2 = 870 votes needed to pass quorum

At the beginning of the cycle this number is entered into a stats table so the clients can pull the data and show it on the UX ("870 needed to pass, this one has 600..." - something like that)

I also have been wanting to have information which proposals were voted on in which cycles so if we can do that would be great. If it's easy enough. It's a bit tricky though as some may have been skipped due to low stake, but I think that was only 1 or 2. 

We can also eventually switch to iterating through proposals by "open" status rather than always checking all proposals which will eventually fail when we have too many. 

stats table
```
      TABLE cycle_stats_table {
        uint64_t propcycle; 
        
        uint64_t start_time; 
        uint64_t end_time; 
        uint64_t num_proposals;
        uint64_t total_voice_cast;
        uint64_t total_favor;
        uint64_t total_against; 
        uint64_t total_citizens; // total number of eligible voters - don't need to backfill this, just use 0 for unknown
        uint64_t quorum_vote_base; // avg votes cast of the last N cycles (N=2)
        uint64_t quorum_votes_needed; // quorum_vote_base * quorum % - for clients to show how much needed.
        float unity_needed; // since unity can be adjusted we want to know what the requirement was

        Vector<uint64_t> active_props[]; // active props up for voting this cycle
        Vector<uint64_t> eval_props[];   // eval phase props up for voting on this cycle

        uint64_t primary_key()const { return propcycle; }

      };

```

## Background
Hey @everyone Hypha has been putting a lot of time, thought and energy into trying to find version 1.0 of the governance system to pass over to the Citizens (where after this proposed change ALL further changes will be completely guided by the governance process by Citizens (and the referendum category will be active). 

We propose the following changes for the next cycle. Is there any blocks or strong objections to these changes? Any thoughts? 

## **Increasing minimum quorum to 7% **
WHY: 
[...]

PR #208, opened 2021-01-12
Quorum is generated from the avg voice cast form the last 2 cycles #207

IanMendozaJaimes:
## Spec 
#207 

### Before moon cycle
- Disable quorum check in code for initial deploy - all current proposals are passing the existing quorum calculation so this does not impact any existing proposal in this cycle. 
- Deploy Proposals, Settings
- Update settings
- Manually set the old unity setting back to 80%
- Call migration action on proposals
- run moon cycle as usual

### After moon cycle
- Re-enable quorum check
- Deploy proposals
- Update settings

## Dev Notes
- Added `cyclestats` and `cycvotedprps` tables
- Added tests

The `cyclestats` table is the table that you described in the issue. The quorum base is being calculated using the value `total_voice_cast` which is the sum of all the voice amount spent during the cycles. I got a bit confused when you say "avg votes cast" that's why I added the field `num_votes` which is the sum of all the number of votes during the cycle. I added a comment in the method `calc_quorum_base` in case the quorum base should be calculated using the total num of votes instead of the voice.

>I also have been wanting to have information which proposals were voted on in which cycles so if we can do that would be great. If it's easy enough. It's a bit tricky though as some may have been skipped due to low stake, but I think that was only 1 or 2.

I added the table `cycvotedprps` scoped by propcycle, to store which proposals have been voted during the cycle. I added the action `migrtevotedp` which walks through the props table and uses the passed_cycle property to determine the cycle the proposal was voted. Unfortunately, I couldn't figure out a way to know in which cycle the rest of the proposals were voted.

> We can also eventually switch to iterating through proposals by "open" status rather than always checking all proposals which will eventually fail when we have too many.

About that, I think the index `by_status` that is in proposals table might not be enough, because when a proposal change status, the whole structure rearranges and then I think it increments the computational complexity. It also starts doing weird things due to the reordering and it's difficult to predict where the pointer will be next, I think that if I try to fix that it will be not as efficient as it should be.
Another way, is to use the new vectors active_props and eval_props but I still need to think how to solve efficiently the staged props, maybe creating another vector of staged proposals before evaluate them. Another solution might be to create a vector of pointers storing the pointers of the props in open status, that vector won't be rearranged but it will be like passing twice over the open and evaluate proposals

### DIFF
--- a/src/seeds.proposals.cpp
+++ b/src/seeds.proposals.cpp
@@ -58,6 +58,11 @@ void proposals::reset() {
     }
   }
 
+  auto citr = cyclestats.begin();
+  while (citr != cyclestats.end()) {
+    citr = cyclestats.erase(citr);
+  }
+
 }
 
 bool proposals::is_enough_stake(asset staked, asset quantity, name fund) {
@@ -271,7 +276,12 @@ void proposals::onperiod() {
     cycle_table c = cycle.get_or_create(get_self(), cycle_table());
     uint64_t current_cycle = c.propcycle;
 
+    std::vector<uint64_t> active_props;
+    std::vector<uint64_t> eval_props;
+
     while (pitr != props.end()) {
+      uint64_t prop_id = pitr -> id;
+
       if (pitr->stage == stage_active) {
         votes_tables votes(get_self(), pitr->id);
         uint64_t voters_number = distance(votes.begin(), votes.end());
@@ -314,6 +324,8 @@ void proposals::onperiod() {
               proposal.current_payout += payout_amount;
             });
 
+            eval_props.push_back(prop_id);
+
           } else {
             
             uint64_t age = pitr -> age + 1;
@@ -334,6 +346,8 @@ void proposals::onperiod() {
                 proposal.executed = true;
                 proposal.status = status_passed;
                 proposal.stage = stage_done;
+              } else {
+                eval_props.push_back(prop_id);
               }
               proposal.current_payout += payout_amount;
             });
@@ -361,11 +375,14 @@ void proposals::onperiod() {
           proposal.stage = stage_active;
         });
         size_change(prop_active_size, 1);
+        active_props.push_back(prop_id);
       }
 
       pitr++;
     }
 
+    update_cycle();
+    update_cycle_stats(active_props, eval_props);
     updatevoices();
     
     transaction trx_erase_participants{};
@@ -378,8 +395,6 @@ void proposals::onperiod() {
     // I don't know how long delay I should use
     // trx_erase_participants.delay_sec = 5;
     trx_erase_participants.send(eosio::current_time_point().sec_since_epoch(), _self);
-
-    update_cycle();
 }
 
 void proposals::updatevoices() {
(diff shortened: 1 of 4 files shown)


IanMendozaJaimes:
> About that, I think the index by_status that is in proposals table might not be enough, because when a proposal change status, the whole structure rearranges and then I think it increments the computational complexity. It also starts doing weird things due to the reordering and it's difficult to predict where the pointer will be next, I think that if I try to fix that it will be not as efficient as it should be.
Another way, is to use the new vectors active_props and eval_props but I still need to think how to solve efficiently the staged props, maybe creating another vector of staged proposals before evaluate them. Another solution might be to create a vector of pointers storing the pointers of the props in open status, that vector won't be rearranged but it will be like passing twice over the open and evaluate proposals

About that, I figured out an easier way. I just needed to use 2 pointers instead of 1. I pushed the change as well.

src/seeds.proposals.cpp
```diff
@@ -271,101 +275,125 @@ void proposals::onperiod() {
...
 
-        double majority = double(prop_majority) / 100.0;
-        double fav = double(pitr->favour);
-        bool passed = pitr->favour > 0 && fav >= double(pitr->favour + pitr->against) * majority;
-        name prop_type = get_type(pitr->fund);
-        bool is_alliance_type = prop_type == alliance_type;
-        bool is_campaign_type = prop_type == campaign_type;
+    if (props_by_status.begin() != props_by_status.end()) {
+      auto p1 = props_by_status.begin();
+      auto p2 = p1;
```
n13, line 283: @IanMendozaJaimes Are you sure this is going to work - that p2 won't be changed when p1 changes due to changes to status in the props list?

How I would solve this is to iterate over status eval and capture the IDs, then iterate over status open and capture those IDs, then iterate over the list of Ids captured and find proposals by ID. 

Less efficient, admittedly, but no pointer magic. Anwyay we shall see I hope it works like this ;)

src/seeds.proposals.cpp
```diff
@@ -271,101 +275,125 @@ void proposals::onperiod() {
...
-        size_change(prop_active_size, -1);
-      
-      }
-
-      if (pitr->stage == stage_staged && is_enough_stake(pitr->staked, pitr->quantity, pitr->fund) ) {
-        props.modify(pitr, _self, [&](auto& proposal) {
-          proposal.stage = stage_active;
-        });
-        size_change(prop_active_size, 1);
+        p1 = p2;
```
n13, line 386: Hmm... before you did pitr = p1, but running under the assumptin that changes to pitr will also cause changes to p1

now you do p1 = p2 - would we then not assume that changes to p1 cause changes to p2 as well? Meaning it's a pointer object rather than a copy.

IanMendozaJaimes: Well, it shouldn't be modified, because they are two different variables, so after I assign p1 = p2, I do p2++, now p2 is pointing to the next value. If p1 changes status, then p1 --> next() would point to something else because of the rearrangement, but p2 keeps pointing to the same value which is the value p1 now takes, and then p2++ points to the next value and so on. I must admit this sounds tricky and it's easy to make a mistake, I can just tell it passes the proposals' tests but sometimes not everything shows up there. So yeah it might be to risky to try

PR #214, opened 2021-01-23
Call Harvest from the scheduler #212

IanMendozaJaimes:
- Added `runharvest` to scheduler `reset`
- Added permissions
- Added tests in scheduler
- Modified the `runharvest` action so it won't crash in case mint rate is empty


n13 (approved):

n13:
### Deployment

1. deploy harvest
2. updatePermissions
3. updateScheduler

PR #221, opened 2021-02-06
Upgrade EOSjs #211

IanMendozaJaimes:
- Added eosjs port

The eosjs-port, creates an interface for the eosjs methods that were used throughout the scritps. I added it int he helper.js so in theory everything else should remain the same. I just modified a bit some other scripts (but I made sure they still work).


scripts/deploy.js
```diff
@@ -464,6 +515,8 @@ const deployAllContracts = async () => {
     if (account.quantity && Number.parseFloat(account.quantity) > 0) {
       await transferCoins(accounts.token, account)
     }
+
+    await sleep(1000)
```
juliolrmonteiro, line 519: Why do you need sleep here?

scripts/stakeusers.js
```diff
@@ -18,15 +18,36 @@ const main = async () => {
...
+
+      // const { transaction_id } = await eos.transaction(async trx => {
+      //   await trx.delegatebw({
+      //     from: 'seedsharvest',
+      //     receiver: user.account,
+      //     stake_net_quantity: '0.1000 TLOS',
+      //     stake_cpu_quantity: '2.0000 TLOS',
+      //     transfer: 0
+      //   }, { authorization: `seedsharvest@owner` })
+      // })
```
juliolrmonteiro, line 50: Is it really necessary to leave this comment in?

juliolrmonteiro (approved):
Looks good to me.

n13 (approved):
tested deployment, added some error message improvements, looks great

PR #229, opened 2021-02-10
DRAFT cycle stats based on proposal types

n13:
cycle stats needs to be separated out for each type of proposal


n13:
Tried this, but didn't make a difference and actually had a negative impact.

This is a lot of work and not sure it makes sense to do.

Will be fixed by using the current deployed vote power and comparing to that

Issue #231, opened 2021-02-17
Flagging Feature


n13:
Action: Flag another user

Flag points - same as rep points calculations

Resident has 10
Citizen has 20

x rep multiplier, so for example rep multiplier is 1.8, x 20 base points = 38 points for a citizen in the 90% percentile

Once a user has 100 flag points, they get the "punish" action triggered on them, which ripples negative reputation out to all who vouched for them

## Spec

Add an action "flag" (from_user, to_user)
Add a table flag_table
Flag_table keeps flag points, scoped by receiving user, list of users who did the flagging

User can call flag on another user only once
User can remove their flag on another user
Flag points are checked any time there is a new flag incoming
When the flag threshold is triggered the punishment is executed - it cannot be undone since it ripples out
Flag threshold is a setting of high impact "flag.thresh"

Constitution & Game guide pp 53

## Implementation - 

Possible to continue to remove reputation of the target after the 100 pt threshold.

Try to do it like this:
- Accumulate up to 100 flag points (threshold) without taking action
- Once 100 is reached, we take away these 100 and any extra points that get in there
- Once 100 is reached, additional flags from people directly remove reputation in the same amount

Example
A flags Z for 40 points - nothing happens, but it is stored in the flag table
B flags Z for 35 points - nothing happens
C flags Z for 30 points - that's 105 total so A gets punished for 105 points, and they ripple out to vouchers
D flags Z for 20 points - A gets punished for an additional 20 points, which also ripples out
... and so on

The idea is if a scammer managed to get a lot of reputation, this way it can be removed again

juliolrmonteiro:
So each user has their own flagging "power" that can only change when their rep changes. They can flag as many users they want, and each time they flag someone the current full "flagging power" is applied. If they want to "correct" they flagging, they have to remove the past flag and reapply a flag.

- Suppose A(Citizen with 1.5 rep) flagged Y for 30 points
- Then in the future A increased his rep to 2.0.
- User A (Citizen with 2.0 rep) flags user X with 40 points
- If they want to correct their flag on Y, they have to remove the flag first (-30 points) and reapply (+40 points).

IanMendozaJaimes:
Hello I have some questions:
> User can remove their flag on another user
- Does that mean that a user can only flag one user at a time? I mean if user A has flagged user B, then A can not flag user C until A removes her flag from B?
[...]

Issue #226, opened 2021-02-08
Rippling Penalties on Reputation


juliolrmonteiro:
When an account get penalized as a "bot", those that vouched for this account also need to get penalized. And those that vouched on the vouched too (ripples).

Also implement the "bot" penalty.

If possible also implement the failed proposal -100 penalty (this is not supposed to be rippled)

PR #234, opened 2021-02-22
Flagging Feature #231

IanMendozaJaimes:
- Added functions to flag and remove flag from a user
- Modified `punish` function so it can remove variable rep points and punish vouchers using batches
- Added tests

The `punish` action, also punishes the vouchers with a certain percent (50%) of the removed points, this is also mentioned in the issue #226


n13:
## Deployment
-updateSettings
- deploy accounts

include/seeds.accounts.hpp
```diff
@@ -53,7 +55,7 @@ CONTRACT accounts : public contract {
 
       ACTION subrep(name user, uint64_t amount);
 
-      ACTION punish(name account);
+      // ACTION punish(name account);
```
n13, line 58: comment - remove

src/seeds.accounts.cpp
```diff
@@ -976,3 +981,151 @@ uint64_t accounts::rep_score(name user)
...
+
+  flag_points_tables flags(get_self(), to.value);
+  flag_points_tables total_flags(get_self(), flag_total_scope.value);
+  flag_points_tables removed_flags(get_self(), flag_remove_scope.value);
+
+  auto fitr = flags.find(from.value);
+  check(fitr == flags.end(), "a user can only flag another user once");
+
+  auto to_rep = rep.find(to.value);
+  check(to_rep != rep.end(), to.to_string() + " user has no reputation");
```
n13, line 1046: Should fail silently, not fail the transaction? If I flag someone but they don't have a reputation, the fact I flagged them is still OK.

n13 (changes requested):
requested 2 minor changes, otherwise looks fantastic.

IanMendozaJaimes:
- Added requested changes

src/seeds.accounts.cpp
IanMendozaJaimes: Ok, I added a validation. In case the flagged user does not have rep, the total_flag points are still increased, but the total_removed not, so when the user has rep and if the user is flagged again, the user will lose the pending points as well

n13:
One more comment 0 similar, deal with rep 0 of from

n13 (approved):

Issue #233, opened 2021-02-17
Accept Vouch and Un-vouch on penalty


n13:
## Overview

We will have forward penalties where users lose reputation gained from vouching

But in the current system we don't have a way to do that

Currently vouch points are added up and capped at 50

In the new proposed system we remember each vouch, and the points, in a table scoped by user

Then in the places we use these vouch rep points, we count max. of 50. The calculation can be redone when a vouch falls away.

## Spec
Just like penalty points, keep vouch points in a table thats scoped by user
Alter the main rep points of a user when the total changes - I think we can figure this out on each additional vouch

## Penalty rippling forward 

When a user gets penalized, the users this user vouched for lose the vouch points they gained.

Example user A gets penalized
A vouched with 22 points for B
B now loses those 22 points. B's reputation might change from that if they had less than 50 points, or if they had more than 50, but after the penalty now have less than 50. 

B's reputation would not change if they still have >= 50

juliolrmonteiro:
Check Constitution at this point: https://docs.google.com/document/d/1C4w9Ol8VGabCIcQDVPDrwcTRoJXBqhrb7VjslwQbUGU/edit#heading=h.95nxy840131q

These 3 lines are relevant:

Index | Name / Action | Description / Calculation | Setting / | Setting ID | Impact | Contract
-- | -- | -- | -- | -- | -- | --
  | Flag Reputation Loss Limit Before Effect | Members lose reputation when flags reach a minimum of N flag points. This triggers the reputation loss for any account that vouched for the flagged account. | (-) 100 Rep | flag.thresh |   |  
-- | -- | -- | -- | -- | -- | --
  | Flag Base Points | Each flag comes with N negative reputation points | Resident = (-) 10 Citizen = (-) 20 | flag.basept |   |  
-- | -- | -- | -- | -- | -- | --
  | Vouched Flagged Account | Lose reputation if an account you have invited or vouched for is fully flagged | (-) 50 Rep |   |   |  
-- | -- | -- | -- | -- | -- | --

So it seems the voucher account loses 50 points, while the vouched account that got flagged loses 100.

IanMendozaJaimes:
Hello, I have some questions:
- Should the function `unvoche` be implemented as an option for the sponsor?
- Related to the example above, when A is penalized, then B losses her 22 vouch points, but this should be considered like an `unvoche` action or instead, A remains as a sponsor in the vouche table but A gives 0 vouch points now

[...]

PR #235, opened 2021-02-24
Accept Vouch and Un-vouch on penalty #233

IanMendozaJaimes:
- Added a new table for vouches
- Added tests
- Added a migration function

### DIFF
--- a/include/seeds.accounts.hpp
+++ b/include/seeds.accounts.hpp
@@ -21,6 +21,7 @@ CONTRACT accounts : public contract {
           refs(receiver, receiver.value),
           cbs(receiver, receiver.value),
           vouches(receiver, receiver.value),
+          vouchtotals(receiver, receiver.value),
           reqvouch(receiver, receiver.value),
           rep(receiver, receiver.value),
           sizes(receiver, receiver.value),
@@ -148,6 +149,7 @@ CONTRACT accounts : public contract {
       uint32_t num_transactions(name account, uint32_t limit);
       void add_active (name user);
       void add_cbs(name account, int points);
+      void calc_vouch_rep(name account);
 
       DEFINE_USER_TABLE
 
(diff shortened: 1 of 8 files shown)


n13:
## Deployment

- update Accounts contract
- run migration action

n13:
@IanMendozaJaimes onboarding test fails

./scripts/seeds.js test onboarding

    ✖  Given invited new user: should have new vouch entry
    -------------------------------------------------------
        operator: deepEqual
        expected: [0, 1]
        diff: [
          1: 1 => 0
        ]
        source: at /Users/nik/work/seeds/seeds-contracts/test/onboarding.test.js:425:5
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (internal/process/task_queues.js:93:5)

n13 (changes requested):
Please fix onboarding test

Also: I fixed some ecc errors we got because of changing library, please check into it and perhaps remove the other / old "ecc" import, or the "getEcc()" function. 

I noticed that getECC is getting the sha256 function from a different version of the eosio.ecc package, which made it not work

That version returns an array of integers

The new version returns a hex string

The old version didn't work so my quick fix was to fix it in one place to make onboarding and accounts user tests work again (invite function)

Please check ;)

IanMendozaJaimes:
Hello @n13, I added the requested changes:
- I fixed the sha256 function in the helper, now onboarding tests work
- I removed eosjs-ecc from package.json, the eosjs v21 library doesn't use it as eosjs-ecc is deprecated and unmaintained, that's why I preferred to keep using the ecc from the eosjs v21}
- Also, I added the attribute `points` to the `punish` function, it is not used in this issue but I added it because it is used in the flagging issue, so that should cause less problems when merging

n13:
Works perfectly, thank you

n13 (approved):

PR #244, opened 2021-03-06
Fix min trx value

n13:
Minor bugfix - if transactions number from planted is less than minimum, make it minimum

Bug scenario

Minimum transaction = 7
Multiplier = 1x

User has 1 Seeds planted, tx limit would now be 1 in the old code

In the new code, it's never below the minimum.

### DIFF
--- a/test/token.test.js
+++ b/test/token.test.js
@@ -332,3 +332,73 @@ describe('token.resetweekly', async assert => {
 
 })
 
+describe.only('transaction limits', async assert => {
+
+  if (!isLocal()) {
+    console.log("only run unit tests on local - don't reset accounts on mainnet or testnet")
+    return
+  }
+
+  const contracts = await initContracts({ token, settings, accounts, harvest })
+  
+  console.log('accounts reset')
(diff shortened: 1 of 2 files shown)


test/token.test.js
```diff
@@ -332,3 +332,73 @@ describe('token.resetweekly', async assert => {
 
 })
 
+describe.only('transaction limits', async assert => {
```
IanMendozaJaimes, line 335: The test works fine, I just want to point out that the `only` function is doing that this test is the only one being executed

IanMendozaJaimes (approved):
Everything looks great. 
Although, I think the first test in token should be fixed, because it is using the old transactions table

test/token.test.js
n13: Good catch thanks

n13:
Deployed

n13:
> Everything looks great.
> Although, I think the first test in token should be fixed, because it is using the old transactions table

I noticed when I uncommented "only"

I removed the history table tests, since they're done in the history contract tests.

PR #252, opened 2021-03-09
Only residents can create proposals, only citizens can create referen…

n13:
Fux for 

only residents can create proposals

only citizens can create referendums

missing: update unit tests

### DIFF
--- a/test/proposals.test.js
+++ b/test/proposals.test.js
@@ -1839,8 +1839,8 @@ describe('Voice scope', async assert => {
   console.log('join users')
   await contracts.accounts.adduser(firstuser, 'firstuser', 'individual', { authorization: `${accounts}@active` })
   await contracts.accounts.adduser(seconduser, 'seconduser', 'individual', { authorization: `${accounts}@active` })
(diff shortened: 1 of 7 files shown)


test/proposals.test.js
```diff
@@ -47,6 +49,9 @@ describe('Proposals', async assert => {
   await contracts.accounts.adduser(thirduser, 'thirduser', 'individual', { authorization: `${accounts}@active` })
   await contracts.accounts.adduser(fourthuser, 'fourthuser', 'individual', { authorization: `${accounts}@active` })
 
+  await contracts.accounts.testresident(firstuser, { authorization: `${accounts}@active` })
+  await contracts.accounts.testresident(seconduser, { authorization: `${accounts}@active` })
```
IanMendozaJaimes, line 53: I think the `fourthuser` should be included as a resident too, because below this user creates a proposal below, and this test is failing due to that.

IanMendozaJaimes (approved):
This looks great! I would only add the fourth user as a resident in the first test

PR #254, opened 2021-03-09
Bugfix: Nonce added only on local node

IanMendozaJaimes:
Added validation so the nonce action from eosjs-port is only added on local node


n13 (approved):
Looks great!

Issue #236, opened 2021-02-25
Invite Campaign Type


n13:

IanMendozaJaimes:
Hello, I have some questions:
- I think I don't understand the reward attribute, is the value of reward the rep points the reward owner will get after someone accepts an invitation?
- Also, is there a setting that establishes a max reward amount?
- It's possible to send rewards to ourselves as the example states, right?

- In the cancel invite function, the sponsor can cancel it and the seeds go to the sponsor again, but in the case of a campaign, do the seeds go to the sponsor directly, or they go to the campaign instead?

- Should campaigns have an status attribute? (to know if they were completed, canceled, etc)

n13:
> Hello, I have some questions:
> 
> * I think I don't understand the reward attribute, is the value of reward the rep points the reward owner will get after someone accepts an invitation?

No it's a new feature in invite campaign - each invite that is accepted results in an immediate SEEDS reward payout for the beneficiary of the reward (reward owner). 

This needs to be part of the parameters


> * Also, is there a setting that establishes a max reward amount?

Yes, max should be a setting, and should be set to 1,000 initially.

> * It's possible to send rewards to ourselves as the example states, right?

Yes, normally someone let's say useraa, creates the campaign as a proposal. Then they would by default be the reward owner. But it's good to have it flexible so we can for example have useraa create the campaign, but let rewards go to an organization of their choosing (part of the campaign parameters)

> * In the cancel invite function, the sponsor can cancel it and the seeds go to the sponsor again, but in the case of a campaign, do the seeds go to the sponsor directly, or they go to the campaign instead?. Also, the cancel function could be called by any authorized account that created  the invite, right?

Let's say 10,000 Seeds were paid into the balance, 1,000 were used, but then the project is abandoned

Then either the owner (beneficiary) can cancel it

Or it gets downvoted and the proposals contract cancels it (funds.seeds). 

either way the funds would then go back to where they originally came from. Whoever provided the funds, gets them back. This is the "origin" entry in the table - the origin of the funds, and also where they go back to. 

> * Should campaigns have an status attribute? (to know if they were completed, canceled, etc)

I imagine we could just delete them from the list if canceled? Hmm... Not sure. If you think it's valuable. I don't see an immediate use for the flag  but I am not opposed to have it. 

[...]

PR #255, opened 2021-03-10
Onboarding: Support for Invite Campaigns #236

IanMendozaJaimes:
- Added new tables `campaigns` and `campaigns_invites`
- Added tests
- Added new permissions to helper

## Ticket
https://github.com/JoinSEEDS/seeds-smart-contracts/issues/236

## Deployment 

- Onboarding 
- Settings
- updateSettings

### DIFF
--- a/src/seeds.onboarding.cpp
+++ b/src/seeds.onboarding.cpp
@@ -99,7 +99,7 @@ void onboarding::accept_invite(name account, checksum256 invite_secret, string p
 
   auto _invite_secret = invite_secret.extract_as_byte_array();
   checksum256 invite_hash = sha256((const char*)_invite_secret.data(), _invite_secret.size());
-  
+
   checksum256 empty_checksum;
 
   invite_tables invites(get_self(), get_self().value);
@@ -186,6 +186,16 @@ void onboarding::reset() {
   while (iitr != invites.end()) {
     iitr = invites.erase(iitr);
   }
+
+  auto citr = campaigns.begin();
+  while (citr != campaigns.end()) {
+    citr = campaigns.erase(citr);
+  }
+
+  auto csitr = campsponsors.begin();
+  while (csitr != campsponsors.end()) {
+    csitr = campsponsors.erase(csitr);
+  }
 }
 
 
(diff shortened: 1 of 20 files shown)


src/seeds.onboarding.cpp
```diff
@@ -255,6 +278,31 @@ void onboarding::_invite(name sponsor, name referrer, asset transfer_quantity, a
 
   uint64_t key = invites.available_primary_key();
 
+  if (campaign_id != 0) {
+    auto citr = campaigns.find(campaign_id);
+    check(citr != campaigns.end(), "campaign not found");
+
```
n13, line 284: Maybe we should add the future reward amount here too? Otherwise I could create 1000 invites, use up all the funds for that, then run out of rewards, or exceed the balance. 

Ok given that all balances may be mixed up, I think we need to include reward amount here.

n13:
```
total_quantity = total_quantity + citr -> reward_amount; // something like that
```

n13: Example, invite is 20 seeds total, but 10 seeds reward, we need to calculate the total cost of the invite as 30 Seeds, then compare 30 to remaining balance and make sure there's 30 left, then remove 30 already here (?) but not inside the send reward function.... 

Or maybe there's another solution to keep these balances separate. Maybe the rewards are kept track of separately?

src/seeds.onboarding.cpp
```diff
@@ -364,6 +435,205 @@ void onboarding::cleanup(uint64_t start_id, uint64_t max_id, uint64_t batch_size
...
+    itr = campinvites_by_campaigns.erase(itr);
+    count++;
+  }
+
+  auto citr = campaigns.find(campaign_id);
+
+  if (itr != campinvites_by_campaigns.end() && itr->campaign_id == campaign_id) {
+
+    transfer_seeds(citr->origin_account, total_refund, "return campaign funds");
+
```
n13, line 621: nice!

n13 (changes requested):
I am not really sure but I think rewards need to be taken into account every time an invite is created - see comments

n13:
Looks great but one potential issue

src/seeds.onboarding.cpp
IanMendozaJaimes: Yes, you are right, I will correct this

IanMendozaJaimes:
- Added requested changes

n13 (approved):

PR #262, opened 2021-03-10
Invite and Funding campaigns

IanMendozaJaimes:
This pull request assumes that the pull request #255 has been added

## Deployment
Onboarding
Proposals
Settings

updateSettings
updatePermissions

## Migration
This branch is the feature branch, the migration branch is: **migration/invite-funding-campaigns**. The migration methods should be run first.

For the migration: there are two methods:
- `migprops`: migrates from `props` (old version) to a temporal table called `migrateprops`, this method don't erases any entry of `props` (just in case something went wrong)
- The second method is `resetprops`, this method deletes the `props` table

Then the next step is in this branch, a method called `initprops`, that action transfers the data from `migrateprops` to the new version of `props`


src/seeds.proposals.cpp
```diff
@@ -301,6 +328,8 @@ void proposals::onperiod() {
     while (pitr != props.end()) {
       uint64_t prop_id = pitr -> id;
 
+      print("running proposal: ", pitr->id, "\n");
+
       // active proposals are evaluated
```
n13, line 333: I think we need to merge master into this to get the new quorum calculations

src/seeds.proposals.cpp
```diff
@@ -382,6 +419,10 @@ void proposals::onperiod() {
             burn(pitr->staked);
           }
 
+          if (pitr->campaign_type == campaign_invite_type) {
+            send_return_funds_campaign(pitr->campaign_id);
+          }
+
```
n13, line 425: I guess invite campaigns need to stay in evaluate until all their money is spent?!

src/seeds.proposals.cpp
```diff
@@ -826,6 +939,10 @@ void proposals::updatex(uint64_t id, string title, string summary, string descri
   require_auth(pitr->creator);
   check(pitr->favour == 0, "Prop has favor votes - cannot alter proposal once voting has started");
   check(pitr->against == 0, "Prop has against votes - cannot alter proposal once voting has started");
+  
+  if (pitr->campaign_type == campaign_invite_type) {
+    pay_percentages = { 100, 0, 0, 0, 0, 0 };
+  }
```
n13, line 945: Cool!!! I guess adding a few 0 we could get away with it for now

n13 (approved):
Approved with comments

src/seeds.proposals.cpp
IanMendozaJaimes: Yes, I updated it

IanMendozaJaimes:
- Merged master to include the new quorum calculations

n13:
Need to run migration before merging this

Get the migration branch - compile and deploy - run the migration method - then merge this

n13:
Migrations DONE

n13:
Unit tests for onboarding, proposals pass 

1 fail on vote power

n13:
Deployed

PR #264, opened 2021-03-12
update for real quorum calculation, using vote power

n13:
onperiod and testperiod now use vote power calculation for quorum


IanMendozaJaimes (approved):
This looks great!

n13:
Deployed

PR #266, opened 2021-03-15
Unit tests updated for new quorum calculations

n13:
New quorum calculation upset many unit tests, fixing it up so they all pass once again


IanMendozaJaimes (approved):
Every test passed!!

PR #268, opened 2021-03-16
added settings for vouch reputation rewards

n13:
* Setting for ref point rewards added
* Code and tests changed for ref point setting
* Some unused settings removed
* Some additional settings added - mainly for the 5 levels of organizations we now have in the game guide

As per game guide & constitution

## Deployment
- deploy settings, accounts
- updateSettings

### DIFF
--- a/include/seeds.accounts.hpp
+++ b/include/seeds.accounts.hpp
@@ -96,6 +96,10 @@ CONTRACT accounts : public contract {
       const name individual = "individual"_n;
       const name organization = "organisation"_n;
 
+      const name citizen = name("citizen");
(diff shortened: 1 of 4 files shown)


IanMendozaJaimes (approved):
- This looks great to me, I would only remove the `unvouch` test since that action is no longer available

n13 (commented):
> * This looks great to me, I would only remove the `unvouch` test since that action is no longer available

Correct, I fixed this. 

Kinda almost not worth removing unvouch.. but ok ;)

n13:
Deployed

PR #279, opened 2021-03-24
Rename bioregions contract to regions

IanMendozaJaimes:
- Renamed bioregions contract to regions
- Renamed methods and settings that included a reference to "bio"
- Added new index in the new regions table
- Changed the name of some fields in the regions contract
- Changed the domain ".bdc" to ".rdc"

I think no migration is needed because those tables were empty by the time I created this pull request.

There was an account in `helper.js` called "bdc", I changed to "rdc", that would require the creation of a new account.

I didn't change the contract's name on the nets, so it is still called "bio.seeds".

#### contracts
- region / bioregion
- onboarding
- history
- harvest
- scheduler
- settings

#### migrations
- updateSettings
- updateScheduler

Unit tests work in this branch
Unit tests work in merged master branch

Issue #248, opened 2021-03-09
Activate Region when 144 citizens have joined


n13:
When 144 citizens have joined a region (formerly bioregion), it becomes active

For now we can simply implement this by
- Counting the number of members in a region - we might already do that
- having an active flag that gets set to true when they've joined
- Have an index for the active flag so we can iterate over only the active regions - only active regions get part of the harvest.

As harvest is concerned, I think we will simply have buckets, then later distribute the buckets to their recipients

IanMendozaJaimes:
- Hello I have a question, for this implementation should I use a migration for the bioregions table? or the active flag could be in another table

n13:
I am ok having it in another table, or even a badge

I have been thinking about badges

n13:
Adding this to milestone as we need it up much sooner than expected

PR #280, opened 2021-03-25
Feature/active regions #248

IanMendozaJaimes:
**This pull request assumes #279 is merged**

- Regions table had a status field, so I used it to determine if the region is inactive/active
- Added the change in the status

Implements issue #248 

## Deploy
Region
Settings
updateSettings


n13:
Deployd ;)

Issue #250, opened 2021-03-09
Active Regions receive part of harvest


n13:
Feature: Only active regions receive a part of the harvest

Inactive don't.

==

From the game guide:

Regional Harvest: Active Regions receive a share of the Harvest. 

This Harvest is managed through a B-DHO (Regional - Decentralized Holonic Organisation) by all the Citizens of that Region. 

Citizens of that region get the benefit of collectively deciding how these funds are spent. See Section 4 Regions for Harvest details and Section 3 for Regional Governance

PR #281, opened 2021-03-25
Active regions receive part of the harvest #250

IanMendozaJaimes:
**- This pull request assumes #280 is merged**

- Added buckets for the regions, now the tokens received from harvest don't go directly to the region account. The buckets use the contract's scope for "SEEDS" and the scope "test" for "TESTS"
- Added tests
- Modified harvest distribution for regions

Implements issue #250 

## Deployment
- harvest 
- region
- settings
- onboarding (was not changed by this PR but want to make sure the latest is actually deployed)

updateSettings


n13 (approved):

n13 (approved):

n13:
Deployed

PR #282, opened 2021-03-25
Bugfix/fix max supply

n13:
Fix max supply, which we call initial supply, to -1


n13:
Deployed

n13:
Won't be deployed, the contract was set, action executed, and set back to master branch

Issue #242, opened 2021-03-04
5 levels of organizations


n13:
## Spec

This is a change to the constitution - we need 5 levels of orgs, not just 3. 

Orgs have 5 levels of sustainability scores, instead of three
Regular, Reputable, Sustainable, Regenerative, Thrivable

There are certain requirements and benefits to each level

pp. 62 of the constitution

![image](https://user-images.githubusercontent.com/65412/109910138-69897000-7ce2-11eb-915f-0a8ead1d73fa.jpeg)

n13:
This method neesd to reflect 5 multipliers

```
double history::get_transaction_multiplier (name account, name other) {
  double multiplier = utils::get_rep_multiplier(account);
  
  auto oitr = organizations.find(account.value);
  if (oitr != organizations.end() && oitr -> status == regenerative_org) {
    multiplier *= config_float_get("regen.mul"_n);
    // Add more multipliers for org
  }

  auto bitr_account = members.find(account.value);
  auto bitr_other = members.find(other.value);

  if (
    bitr_account != members.end() && 
    bitr_other != members.end() && 
    bitr_account -> bioregion == bitr_other -> bioregion
  ) {
    multiplier *= config_float_get("local.mul"_n);
  }

  return multiplier;
}
```

And other changes...

IanMendozaJaimes:
Hello, I have some questions:
1.- Currently the status variable starts from 0 (Regular org), do you want it to start from 1?, a migration would be needed but not to change the table structure, only to add +1 to all the existing orgs.
2.-  There is a table in the history contract that stores when an org becomes reputable and one that stores when an org becomes regenerative, so I think it'll be better to have just one table scoped by type, what do yo think?

juliolrmonteiro:
Regarding (1), why do you think it's best to have status starting at 1?

Regarding (2), yes makes sense to have a single table scoped by type.

Ah, this would be so much easier with Badges :D

IanMendozaJaimes:
(1), I actually think is better to start from 0, in that way is cleaner to map their status in an array, it was just more like an option in case someone wants it to start from 1, but if not, I prefer it from 0
(2) great!

n13:
Starting from 0 is always better in programming :) 

While starting from 1 always introduces issues. 

So I agree with @juliolrmonteiro and @IanMendozaJaimes here

PR #291, opened 2021-04-01
Feature/five level of organizations #242

IanMendozaJaimes:
- Added new methods for orgs to update their status
- Changed the way validations were made for org status
- Added new settings
- Added new table to history contract and deleting `reputable` and `regen` tables, migration is not needed since those tables were empty
- Changed transactions multiplier in history
- Fixed rep multiplier in `addregen` and `subregen` actions
- Added tests for orgs

#242 

## Deployment

**Contracts**
- [ ] Accounts
- [ ] Organization
- [ ] History
- [ ] Settings

updateSettings (yes)

updateScheduler, not needed
migration, not needed

### DIFF
--- a/src/seeds.accounts.cpp
+++ b/src/seeds.accounts.cpp
@@ -1,4 +1,3 @@
- 
 #include <seeds.accounts.hpp>
 #include <eosio/system.hpp>
 #include <eosio/symbol.hpp>
--- a/src/seeds.accounts.cpp
+++ b/src/seeds.accounts.cpp
@@ -202,7 +202,7 @@ void accounts::pnishvouched (name sponsor, uint64_t start_account) {
   require_auth(get_self());
 
   uint64_t batch_size = config_get("batchsize"_n);
-  uint128_t id = uint128_t(sponsor.value) << 64;
+  uint128_t id = (uint128_t(sponsor.value) << 64) + start_account;
 
   auto vouches_by_account = vouches.get_index<"byaccount"_n>();
   auto vouches_by_sponsor_account = vouches.get_index<"byspnsoracct"_n>();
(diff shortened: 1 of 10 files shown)


src/seeds.accounts.cpp
```diff
@@ -203,7 +202,7 @@ void accounts::pnishvouched (name sponsor, uint64_t start_account) {
   require_auth(get_self());
 
   uint64_t batch_size = config_get("batchsize"_n);
-  uint128_t id = uint128_t(sponsor.value) << 64;
+  uint128_t id = (uint128_t(sponsor.value) << 64) + start_account;
```
n13, line 205: Was this a bug you found along the way? Looks like it...

src/seeds.history.cpp
```diff
@@ -80,42 +89,27 @@ void history::addcitizen(name account) {
...
-  reputables.emplace(_self, [&](auto & org){
-    org.id = reputables.available_primary_key();
-    org.organization = organization;
-    org.timestamp = eosio::current_time_point().sec_since_epoch();
+  account_status_tables acctstatus(get_self(), scope.value);
+  
+  acctstatus.emplace(_self, [&](auto & item){
+    item.id = acctstatus.available_primary_key();
+    item.account = account;
+    item.timestamp = eosio::current_time_point().sec_since_epoch();
```
n13, line 100: @IanMendozaJaimes don't you have to remove it from the old scope when you add it to the new scope?

n13: Oh this just records the time when an org became (status) - cool! No questions!

n13 (approved):
Looks great - cleaner than before

PR #297, opened 2021-04-09
Cleanup fix up for campaigns

n13:
Preparing to have an automated refund scheme after 1 cycle, so old invites get refunded

It will use this method to  clean up index ranges

### DIFF
--- a/src/seeds.onboarding.cpp
+++ b/src/seeds.onboarding.cpp
@@ -319,6 +319,10 @@ void onboarding::_invite(name sponsor, name referrer, asset transfer_quantity, a
 
 void onboarding::cancel(name sponsor, checksum256 invite_hash) {
   require_auth(sponsor);
+  _cancel(sponsor, invite_hash, true);
+}
+
+void onboarding::_cancel(name sponsor, checksum256 invite_hash, bool check_auth) {
 
   checksum256 empty_checksum;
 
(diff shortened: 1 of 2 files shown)


IanMendozaJaimes (changes requested):
In case of campaign invites, they should return their funds to the campaign before delete the invite

n13:
> I think in case of campaign invites, they should return their funds to the campaign before delete the invite

Makes sense yes

n13:
> I think in case of campaign invites, they should return their funds to the campaign before delete the invite

I am now sharing the code with cleanup, since it's so complicated - should work!

src/seeds.onboarding.cpp
```diff
@@ -399,14 +405,11 @@ void onboarding::cleanup(uint64_t start_id, uint64_t max_id, uint64_t batch_size
...
-      iitr = invites.erase(iitr);
-      auto refitr = referrers.find(iitr->invite_id);
-      if (refitr != referrers.end()) {
-        referrers.erase(refitr);
-      }
-      count += 4;
+      name sponsor = iitr->sponsor;
+      checksum256 hash = iitr->invite_hash;
+      iitr++;
+      count += 8;
```
n13, line 411: +8 because that's a heavyweight operation there ;)

IanMendozaJaimes (approved):
Looks great!

PR #299, opened 2021-04-15
Feature/gratitude acknowledge

juliolrmonteiro:
This implements the `acknowledge` ACTION to the contract, that allows gratitude users to mark someone to receive a percentage of their tokens without specifying exactly how much (will be decided only when the round ends, based on the number of people they acknowledged).

There's a new setting called `gratz.acks` to indicate how many acknowledges an account has to send before they start diving the remaining tokens by the amount of acks. Default value is 10, so it means if an account does less than 10 acks they will not use up all their tokens for that round, as each ack will use 1/10 of the tokens. If they do more than 10, for example 12, each ack will receive 1/12 of the tokens.

### DIFF
--- a/src/seeds.gratitude.cpp
+++ b/src/seeds.gratitude.cpp
@@ -36,10 +36,11 @@ ACTION gratitude::reset () {
 ACTION gratitude::give (name from, name to, asset quantity, string memo) {
   require_auth(from);
 
-  check( from != to, "gratitude: cannot give to self" );
-  check( is_account( to ), "gratitude: to account does not exist");
+  check(from != to, "gratitude: cannot give to self");
+  check(is_account(to), "gratitude: to account does not exist");
+  check_user(to);
   check_asset(quantity);
-  check( quantity.amount > 0, "gratitude: must give positive quantity" );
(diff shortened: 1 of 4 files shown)


src/seeds.gratitude.cpp
```diff
@@ -46,21 +51,53 @@ ACTION gratitude::give (name from, name to, asset quantity, string memo) {
   update_stats(from, to, quantity);
 }
 
+ACTION gratitude::acknowledge (name from, name to, string memo) {
+  require_auth(from);
+
+  check( from != to, "gratitude: cannot give to self" );
+  check( is_account( to ), "gratitude: to account does not exist");
```
IanMendozaJaimes, line 58: Just a doubt, the "to" user doesn't have to be a seeds user?

IanMendozaJaimes (approved):
Look great!

src/seeds.gratitude.cpp
juliolrmonteiro: Done, added a check_user()

n13 (approved):
GTG

PR #300, opened 2021-04-16
Bugfix/alliance and campaign quorum

n13:
# Code Changes and Cleanup

## Cleanup
Removed various migration data structures and actions, added a clean action, which I already ran on mainnet - clean removed the data from mainnet already. 

## Quorum / Support Level fix for alliances and campaigns

New table: support, scoped by alliance and campaign

This table contains number of proposals, number of votes, and resulting support level needed - all this used to be in cycle stats. 

Cycle stats are not scoped and now have 0 for these values - I wonder if we can remove them without migration? Not sure so leaving them there, will also prevent apps from crashing.

Using support table when adding proposals, adding votes, and when doing the evaluation

## Unit Tests
Cleaned up unit tests and added some for the new support table

```
  passed: 101,  failed: 0  of 101 tests  (3m 15.7s)
```

### DIFF
--- a/scripts/helper.js
+++ b/scripts/helper.js
@@ -29,7 +29,7 @@ const endpoints = {
   local: 'http://127.0.0.1:8888',
   kylin: 'http://kylin.fn.eosbixin.com',
   telosTestnet: 'https://test.hypha.earth',
-  telosMainnet: 'https://node.hypha.earth'
+  telosMainnet: 'https://api.telosfoundation.io'
 }
 
 const ownerAccounts = {
--- a/src/seeds.proposals.cpp
+++ b/src/seeds.proposals.cpp
@@ -1689,9 +1689,12 @@ void proposals::recover_voice(name account) {
 
 }
 
-void gratitude::size_change(name id, int delta) {
+void proposals::size_change(name id, int64_t delta) {
+  size_tables sizes(get_self(), get_self().value);
+
   auto sitr = sizes.find(id.value);
   if (sitr == sizes.end()) {
+    check(delta >= 0, "can't add negagtive size");
     sizes.emplace(_self, [&](auto& item) {
       item.id = id;
       item.size = delta;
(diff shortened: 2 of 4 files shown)


scripts/helper.js
```diff
@@ -29,7 +29,7 @@ const endpoints = {
   local: 'http://127.0.0.1:8888',
   kylin: 'http://kylin.fn.eosbixin.com',
   telosTestnet: 'https://test.hypha.earth',
-  telosMainnet: 'https://node.hypha.earth'
+  telosMainnet: 'https://api.telosfoundation.io'
```
n13, line 32: @sebastianmontero I had timeouts trying to deploy proposals so had to use another node - not sure why our node times out on deploy, or what makes it time out vs. other contracts deployed?

src/seeds.proposals.cpp
```diff
@@ -249,6 +262,91 @@ void proposals::calcvotepow() {
...
+  check(sitr != support.end(), "cycle not found "+std::to_string(cycle));
+  support.modify(sitr, _self, [&](auto & item){
+    uint64_t total_voice = item.total_voice_cast + voice_cast;
+    item.total_voice_cast = total_voice;
+    item.voice_needed = calc_voice_needed(total_voice, item.num_proposals);
+  });
+}
+
+uint64_t proposals::calc_voice_needed(uint64_t total_voice, uint64_t num_proposals) {
+  return ceil(total_voice * (get_quorum(num_proposals) / 100.0));
```
n13, line 338: There's one small bugfix here - using ceil() of this value

Example if 19.2 votes are required at minimum, then really, 20 votes are required, as 19 < 19.2 and 20 > 19.2

n13:
Results
```
❯ cleosm get table funds.seeds campaign support

{
  "rows": [{
      "propcycle": 32,
      "num_proposals": 18,
      "total_voice_cast": 7179,
      "voice_needed": 503
    },{
      "propcycle": 33,
      "num_proposals": 7,
      "total_voice_cast": 3991,
      "voice_needed": 559
    }
  ],
  "more": false,
  "next_key": ""
}

~/work/seeds/seeds-contracts bugfix/alliance_and_campaign_quorum*
❯ cleosm get table funds.seeds alliance support

{
  "rows": [{
      "propcycle": 32,
      "num_proposals": 2,
      "total_voice_cast": 5055,
      "voice_needed": 2022
    },{
      "propcycle": 33,
      "num_proposals": 1,
      "total_voice_cast": 1841,
      "voice_needed": 737
    }
  ],
  "more": false,
  "next_key": ""
}


```

IanMendozaJaimes (approved):
It looks great!

n13:
Merged also this has been deployed since it was a hotfix

Issue #316, opened 2021-05-10
Token transfer bug - not all actions are recorded in history


n13:
here
```
void history::trxentry(name from, name to, asset quantity) {
```

On the delayed transaction, we use "from" as ID, and cancel previous from so unit tests work 

However, this will result in transfers not getting recorded in history

Example a transaction could contain 100 transfer actions - in our system, only the last one would fire a delayed action, so this doesn't work.

I think we need to use a random number to guarantee the delayed actions are all created and running.

@IanMendozaJaimes let me know what you think about this

## Unit Test

I think before we do anything we can simply create a unit test with a worst case scenario, 3 transfer actions from the same account in one transactions. 

Expected result: All 3 are recorded

This will fail as it only records one of them in history (this is the bug)

Then we fix the broken test by changing the code ;)

IanMendozaJaimes:
Hello @n13, the problem is how we can generate the random number, because in eosio it is not possible to use rand https://github.com/EOSIO/eos/issues/2643

and the block time might not work if the transaction contains several transfer actions, also to hash the other parameters might not work if two or more transfers have the same values in their parameters.

I think we can do what hypha is doing for their deferred transactions, they have a table that stores the last id used, and each time they schedule a deferred transaction they increase the counter +1

n13:
> Hello @n13, the problem is how we can generate the random number, because in eosio it is not possible to use rand [EOSIO/eos#2643](https://github.com/EOSIO/eos/issues/2643)
> 
> and the block time might not work if the transaction contains several transfer actions, also to hash the other parameters might not work if two or more transfers have the same values in their parameters.
> 
> I think we can do what hypha is doing for their deferred transactions, they have a table that stores the last id used, and each time they schedule a deferred transaction they increase the counter +1

That sounds like a perfect solution...

Issue #315, opened 2021-05-10
Make Token transfer more efficient


n13:
## Overview

Part (1) of #314 

Currently token transfer accesses the users table, loading entire user objects, 4 times

2 times in the token contract and 2 times in history contract on 
```
void history::trxentry(name from, name to, asset quantity) 
```
## Token contract
This is easy to remove, we don't need user objects we only need planted scores. We can use planted balances to determine if someone is a user - all users have an entry - and handle the special harvst.seeds entry in planted (ignore, not a user).

## History contract

The latter more complicated as we are checking for organization status of an account - we could put that in the delayed transaction

I also found a bug with history entries and delayed transactions, made another ticket for that here: https://github.com/JoinSEEDS/seeds-smart-contracts/issues/316

Overall the strategy should be to keep the call into history as an inline action, as we may need this for other purposes, and we don't want to change the token contract, but perhaps change history contract to delay computationally expensive tasks into a deferred transactions. 

Or, if we clean up the users table and restrict the size of the user object, this won't be an issue ;)

IanMendozaJaimes:
Hello @n13, I think the most expensive part in history::trxentry is to check whether the accounts are orgs or individuals and the get_rep_multiplier function as it checks the orgs table and the regions table.

Maybe what I can do is to keep the part that saves the totals (from line 195 to line 228) and leave all the daily_transactions_tables stuff in a deferred transaction, that includes the calculation of points the transaction receives. Although the checking if an account is an org would be still needed in the normal action, but I can use the orgs table instead of the users table. Is that ok?

I also think this won't be an issue if we can restrict the user object's size

n13:
I think we should use orgs table instead of users table - basically avoid users table whenever we can in the token transfer function.

Not sure we need the deferred tx, that would work but maybe let's do the migration first

n13:
> Hello @n13, I think the most expensive part in history::trxentry is to check whether the accounts are orgs or individuals and the get_rep_multiplier function as it checks the orgs table and the regions table.
> 
> Maybe what I can do is to keep the part that saves the totals (from line 195 to line 228) and leave all the daily_transactions_tables stuff in a deferred transaction, that includes the calculation of points the transaction receives. Although the checking if an account is an org would be still needed in the normal action, but I can use the orgs table instead of the users table. Is that ok?
> 
> I also think this won't be an issue if we can restrict the user object's size

That sounds good to me. Makes sense to have all calculations in the deferred tx. 

[...]

PR #322, opened 2021-05-11
Make Token transfer more efficient #315 (History part)

IanMendozaJaimes:
- Removed the trx points calculation from the main token::transfer action.
- Added new table (singleton) to store the ids for deferred transactions.
- Modified tests for harvest, accounts, organizations and history

Considerations:
- The trxentry method now relies on the fact that all the users and orgs have an entry in the planted table from harvest
- Although the deferred transaction for saving the trx points is unique know, it is not ready to process several of calls for the same `from` and the same `to` in the same second, maybe due to the new size of the function. It means this PR doesn't implements a solution for #316 

__Contracts changed__:
- [ ] History

There aren't any other deployment considerations.

### DIFF
--- a/src/seeds.history.cpp
+++ b/src/seeds.history.cpp
@@ -152,45 +152,15 @@ void history::trxentry(name from, name to, asset quantity) {
     return;
   }
 
-  auto from_user = users.find(from.value);
-  auto to_user = users.find(to.value);
-  
-  if (from_user == users.end() || to_user == users.end()) {
+  auto from_planted = planted.find(from.value);
+  auto to_planted = planted.find(to.value);
+
+  if (from_planted == planted.end() || to_planted == planted.end()) {
     return;
   }
 
-
-  uint64_t day = utils::get_beginning_of_day_in_seconds();
-  daily_transactions_tables transactions(get_self(), day);
-
-  uint64_t transaction_id = transactions.available_primary_key();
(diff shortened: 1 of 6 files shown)


src/seeds.history.cpp
```diff
@@ -152,45 +152,15 @@ void history::trxentry(name from, name to, asset quantity) {
...
-    transaction.from = from;
-    transaction.to = to;
-    transaction.volume = quantity.amount;
-    transaction.qualifying_volume = std::min(transactions_cap, quantity.amount);
-    transaction.from_points = uint64_t(ceil(from_capped_amount * get_transaction_multiplier(to, from)));
-    transaction.to_points = to_is_organization ? uint64_t(ceil(to_capped_amount * get_transaction_multiplier(from, to))) : 0;
-    transaction.timestamp = timestamp;
-  });
+  bool from_is_organization = organizations.find(from.value) != organizations.end();
+  bool to_is_organization = organizations.find(to.value) != organizations.end();
```
n13, line 163: The weird thing is organizations keep their "fixed" planted score not in the planted table but rather in their own table ("planted")

This planted value is really the org fee for  a Ssetting up the org, I think
Anyway to figure out if a user is a Seeds user, use both the planted and org tables if one of them is != end() then it's a Seeds user.

We can make this much more lightweight by passing down the info from the token contract, which already checks those things?! Could be in the parameter for trxentry.

PR #323, opened 2021-05-12
Bugfix/token transfer cpu time

n13:
Fix for #315 

Reduce access to users table

Added access to org and planted tables: This is the minimal data we need to figure out if someone is within their limit.

If we wanted to do this in 1 table access, we would need a new planted table.

We need the info on org vs user downstream so we could just add it as parameters to the trx entry, so we don't have to look it up again in history. 

## Update: 4 table lookups is as good as it gets!

Talking with Ian, the history action needs to know the organization's level and also the region, to figure out the multiplier for the transaction points. 

Will look into it more, to add it as delayed action is possible, but Ian ran into some issues with it not always executing. 

We may just pass along the data from org and planted tables to history, so it doesn't need to look it up again. 

It's becoming more clear why transfers are expensive, they do a lot.

### DIFF
--- a/include/seeds.harvest.hpp
+++ b/include/seeds.harvest.hpp
@@ -16,6 +16,7 @@
 #include <tables/cbs_table.hpp>
 #include <tables/cspoints_table.hpp>
 #include <tables/organization_table.hpp>
+#include <tables/planted_table.hpp>
 #include <eosio/singleton.hpp>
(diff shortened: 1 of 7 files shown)


n13:
close for now

n13:
Putting this on ice for now since we didn't have the issue anymore, and it's pretty complex

PR #325, opened 2021-05-12
Limit sizes of String in Account #317

### LINKED ISSUE #317: Limit sizes of String in Account, elsewhere?
All sizes need to be limited on both create and update, so we don't get very large strings in our user objects

Part (2) of https://github.com/JoinSEEDS/seeds-smart-contracts/issues/314

Suggested limits below - 64 for nickname, 512 for lists, 2048 for story, and 512 for image URLs

Keep in mind that the offending account which caused timeouts had **134,000** characters in a string ;) 

```
#define DEFINE_USER_TABLE TABLE user_table { \
        name account; \
        name status; \
        name type; \
        string nickname; \  ====> Limit 64
        string image; \ == Limit 512 (currently URLS have < 200 characters in our AWS buckets, theoretical max is 2048, 
        string story; => Limit 2048
        string roles; \ Limit 512
        string skills; \ Limit 512
        string interests; \ Limit 512
        uint64_t reputation; \
        uint64_t timestamp; \
```

URL length: 
I did a quick check out current URLs are < 200 characters. We limit to 512 to cover all cases, also URLs have a theoretical max of 2048 in most browsers, also they are overall generated and users don't manually invent them, so they will be determined by systems, rather than user-entered. 

I checked Organization table, there's no Strings in there.

A better data architecture would be to move Strings out of the users table completely, and store that data separately.

When we designed this data structure, we did not know that CPU time was such a big factor, or that loading large objects would cause timeouts.

IanMendozaJaimes:
Fixes #317 

- Added validations in accounts::adduser and accounts::update to enforce the new limits for strings

The new limits  are:

string nickname; \  ====> Limit 64
string image; \ == Limit 512 
string story; => Limit 7000
string roles; \ Limit 512
string skills; \ Limit 512
string interests; \ Limit 512

Only the accounts contract was modified.
This modification doesn't affect the way onboarding works

### DIFF
--- a/src/seeds.accounts.cpp
+++ b/src/seeds.accounts.cpp
@@ -103,6 +103,7 @@ void accounts::adduser(name account, string nickname, name type)
   check(is_account(account), "no account");
 
   check(type == individual|| type == organization, "Invalid type: "+type.to_string()+" type must be either 'individual' or 'organisation'");
+  check(nickname.size() <= 64, "nickname must be less than 65 characters long");
 
   auto uitr = users.find(account.value);
   check(uitr == users.end(), "existing user");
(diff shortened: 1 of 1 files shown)


src/seeds.accounts.cpp
```diff
@@ -103,6 +103,7 @@ void accounts::adduser(name account, string nickname, name type)
   check(is_account(account), "no account");
 
   check(type == individual|| type == organization, "Invalid type: "+type.to_string()+" type must be either 'individual' or 'organisation'");
+  check(nickname.size() <= 64, "nickname must be less than 65 characters long");
```
n13, line 107: Would be nice to have the hard coded number only in 1 place, but ok ;)

n13 (approved):
Thank you!! Perfect

n13:
Deployed!

Issue #239, opened 2021-02-26
Add moon phase to scheduler


n13:
## Overview

We need to be able to trigger scheduler actions at moon phases

## Spec 

create a new action to add an operation on a certain moon phase frequency

We have 4 moon phases 

Add another table "moonops"

moonops table has "quarter_moon_cycles" field which is in quarter moon cycles, 1 = 1/4 moon, 4 = 1 full moon cycle

add new create action with parameters
quarter_moon_cycles
```
ACTION configmoonop(name id, name action, name contract, uint64_t quarter_moon_cycles, uint64_t starttime);
```
Not sure if we should enforce that starttime is a table entry in the moon table - we probably should.

Because then it specified exactly for example "start at full moon", and then quarter_moon_cycles = 4 would mean "every full moon"

## Implementation

moonops executes after the regular ops if has_executed is false, but works the same way, sorted by oldest.

when a moon op is executed timestamp is not set to "now" time but instead is set to the official moon time (which may be off by a few seconds), so we know which cycle it last executed.

PR #326, opened 2021-05-18
Add moon phase to scheduler #239

IanMendozaJaimes:
- Added moonops table
- Added `configmoonop` action and modified `removeop` and `pauseop` to interact with the moon operations
- Removed action `confirm` this action wasn't helping to confirm the correct execution, it was just consuming cpu time

Modified contracts:
- [ ] Scheduler

No migrations nor permissions update is needed

### DIFF
--- a/include/seeds.scheduler.hpp
+++ b/include/seeds.scheduler.hpp
@@ -30,11 +30,11 @@ CONTRACT scheduler : public contract {
         // specify start time any time in the future, or use 0 for "now"
         ACTION configop(name id, name action, name contract, uint64_t period, uint64_t starttime);
 
+        ACTION configmoonop(name id, name action, name contract, uint64_t quarter_moon_cycles, uint64_t starttime);
+
         ACTION removeop(name id);
 
         ACTION pauseop(name id, uint8_t pause);
-
-        ACTION confirm(name operation);
         
         ACTION stop();
         
@@ -111,7 +111,7 @@ CONTRACT scheduler : public contract {
 
         typedef eosio::multi_index <"moonops"_n, moon_ops_table,
             indexed_by<"bylastcycle"_n,
-            const_mem_fun<moon_ops_table, uint64_t, &moon_ops_table::by_last_cycle>
+            const_mem_fun<moon_ops_table, uint64_t, &moon_ops_table::by_last_cycle>>
         > moon_ops_tables;
 
         typedef eosio::multi_index <"test"_n, test_table> test_tables;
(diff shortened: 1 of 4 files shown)


include/seeds.scheduler.hpp
```diff
@@ -89,11 +103,17 @@ CONTRACT scheduler : public contract {
         DEFINE_CONFIG_TABLE_MULTI_INDEX
 
         typedef eosio::multi_index < "operations"_n, operations_table,
-            indexed_by<"bytimestamp"_n, const_mem_fun<operations_table, uint64_t, &operations_table::by_timestamp>>
+            indexed_by<"bytimestamp"_n, 
+            const_mem_fun<operations_table, uint64_t, &operations_table::by_timestamp>>
```
n13, line 107: we probably need to do a migration for this table upgrade

We would not need to if we can just reset everything - and maybe that's the best way. 

But there were some ops we don't reset so that the cycles remain intact. 

i.e. example the cycle is 1 week, it fired 3 days ago, 4 days to go

If we now reset this op, it will fire in 7 days instead of 4 days

n13: But, since the table data structure doesn't change - maybe it's good enough to take everything out, and put everything back in, in 1 action.

n13 (approved):

include/seeds.scheduler.hpp
IanMendozaJaimes: Oh sorry, I didn't add a new index for the operations table, I just pressed \n so all the tables were indented in the same way.

src/seeds.scheduler.cpp
```diff
@@ -5,26 +5,49 @@
...
 
-    if (itr == operations.end()) {
-        return false;
-    }
-    if(itr -> pause > 0) {
-        print("transaction " + operation.to_string() + " is paused");
-        return false;
+        if (itr == operations.end()) {
+            return 0;
+        }
```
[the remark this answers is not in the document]
IanMendozaJaimes: it's just safety precaution, but that's true, it can be removed since it the method is not called anywhere else and by the moment we call the operation exists

src/seeds.scheduler.cpp
```diff
@@ -397,20 +457,49 @@ ACTION scheduler::execute() {
...
         }
         itr++;
     }
 
+    if (!has_executed) {
+        auto moonops_by_last_cycle = moonops.get_index<"bylastcycle"_n>();
+        auto mitr = moonops_by_last_cycle.begin();
+        
+        while (mitr != moonops_by_last_cycle.end()) {
+            uint64_t used_timestamp = is_ready_to_execute(mitr->id, "moonop"_n, timestamp);
```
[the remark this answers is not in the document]
IanMendozaJaimes: oh ok, yes I will change it, and yes, it always returns a valid moon cycle id since the start_date must be also a valid id

include/seeds.scheduler.hpp
n13: Oh, ok!

n13:
Merged

n13:
Deployed

PR #327, opened 2021-05-18
Token transfer bug - not all actions are recorded in history #316

IanMendozaJaimes:
- Added a table to store the ids to use them in deferred transactions in history
- Added a new table to store the processed transactions. This is needed for the correct qev and trx points calculation.

Contracts modified:
- [ ] History

No migrations or settings are needed

### DIFF
--- a/include/seeds.history.hpp
+++ b/include/seeds.history.hpp
@@ -50,7 +50,9 @@ CONTRACT history : public contract {
 
         ACTION savepoints(uint64_t id, uint64_t timestamp);
 
-        ACTION sendtrxcbp(name from, name to);
+        ACTION sendtrxcbp(uint64_t deferred_id, name from, name to);
+
+        ACTION updatetxpt(uint64_t deferred_id, name from);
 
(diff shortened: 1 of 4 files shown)


include/seeds.history.hpp
```diff
@@ -239,6 +241,13 @@ CONTRACT history : public contract {
         uint128_t by_account_key() const { return (uint128_t(account.value) << 64) + key.value; }
       };
 
+      TABLE deferred_id_table {
+        uint64_t id;
+      };
```
n13, line 246: can you actually have a table without primary ID?

n13, line 246: Can you have a table without primary key?

IanMendozaJaimes: In this case yes, because it is a singleton. As the table will only contain one value I thought it was a good idea to make it one.

n13 (approved):

n13:
Great stuff!

n13:
Deployed on testnet and mainnet

PR #331, opened 2021-05-25
Separate out Hypha proposal type #330

### LINKED ISSUE #330: Separate out Hypha proposal type
## Issue
The hypha proposal type already exists, but it shares votes with campaigns

## Spec

Add a separate voice category for hypha, let hypha votes come out of that separate pool

This pool is exactly the same as alliance and campaign pools - just a separate pool of votes

This way, hypha proposals do not take away from campaign proposals

IanMendozaJaimes:
- Added a new voice scope: hypha
- Added the seeds.hypha to local accounts for testing purposes

Fixes #330 

- [ ] Proposals

No permissions nor settings update is needed

The new scope will be added automatically in the next call to `onperiod`

To update the support for a given cycle the action `migvotepow` can be used

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -107,7 +107,6 @@ CONTRACT proposals : public contract {
       ACTION testperiod ();
       ACTION testevalprop(uint64_t proposal_id, uint64_t prop_cycle);
 
-      ACTION migvotepow(uint64_t cycle);
       ACTION reevalprop (uint64_t proposal_id, uint64_t prop_cycle);
 
       ACTION testalliance(uint64_t id, name creator, asset quantity, asset current_payout, name status, name stage, name campaign_type);
@@ -432,7 +431,6 @@ extern "C" void apply(uint64_t receiver, uint64_t code, uint64_t action) {
         (calcvotepow)(addcampaign)(checkprop)(doneprop)
         (testperiod)(testevalprop)
         (cleanmig)(testpropquor)
-        (migvotepow)
         (reevalprop)
         (testalliance)(migalliances)
         )
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -142,13 +142,12 @@ CONTRACT proposals : public contract {
       name campaign_type = "campaign"_n;
       name campaign_invite_type = "cmp.invite"_n;
       name campaign_funding_type = "cmp.funding"_n;
-      name campaign_milestone_type = "milestone"_n;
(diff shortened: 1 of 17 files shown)


src/seeds.proposals.cpp
```diff
@@ -524,26 +533,30 @@ void proposals::evalproposal (uint64_t proposal_id, uint64_t prop_cycle) {
...
-        // TODO: if we allow num_cycles == 1, this needs to go into passed instead of evaluate.
-        // uint64_t num_cycles = pitr -> pay_percentages.size() - 1;
+        uint64_t num_cycles = pitr->pay_percentages.size() - 1;
 
         props.modify(pitr, _self, [&](auto & proposal){
           proposal.passed_cycle = prop_cycle;
           proposal.age = 0;
           proposal.staked = asset(0, seeds_symbol);
-          proposal.status = status_evaluate;
+          if (proposal.age >= num_cycles && !is_alliance_type) {
```
n13, line 542: I don't understand this - why !alliance_type... proposal.age == 0 it was just set above...

n13 (changes requested):
Hmm ... 

There's a lot of changes I think I got the specifications wrong here.

So really we only needed to add a new voice scope called milestone and fill this and take milestone votes from there, treat milestone with 100% payout like alliances, and that's it.

include/seeds.proposals.hpp
```diff
@@ -176,7 +182,7 @@ CONTRACT proposals : public contract {
       void recover_voice(name account);
       void demote_citizen(name account);
       uint64_t calculate_decay(uint64_t voice);
-      name get_type (name fund);
+      name get_type (const name & fund);
```
n13, line 185: does this make a difference? just curious...

src/seeds.proposals.cpp
```diff
@@ -268,9 +256,10 @@ ACTION proposals::migvotepow(uint64_t cycle) {
   //auto pitr = props.find(131);// NOTE migration method from cycle 32 onward
```
n13, line 256: We should just delete this migrate action - we don't need it anymore

n13 (approved):
I am OK now - waiting for @IanMendozaJaimes review

include/seeds.proposals.hpp
IanMendozaJaimes: amm no, this small change doesn't make any difference in the overall speed, I mean to use & in the functions makes them faster but that change won't make the difference

IanMendozaJaimes:
Looks great to me

n13:
Deployed!

PR #338, opened 2021-06-02
Create account only on active

IanMendozaJaimes:
- Added action `createacct` so the founder can create the region account when the regions is active
- Fixed tests

- [ ] Regions

No migrations nor settings update are needed


n13 (approved):
looks good!

n13:
Deployed

Issue #320, opened 2021-05-11
Organization uses balance instead of planted score for tx limits


n13:
This is actually really important

Organizations have paid a fee to create an org, that counts as their planted score.

We can either - plant this fee for the org, and make sure orgs can't unplant this initial amount

Or we can add special code that treats that balance as planted

IanMendozaJaimes:
Hello @n13, I have some questions:
- Currently when an organization is created, the fee they paid is automatically planted in harvest, so I think the easiest thing to do is just not let them unplant that amount, but the amount the orgs aren't allowed to unplant should be only the fee they pay when the are created or the planted amount they should have depending on their status? I mean an org with status `regular` is created with 200 Seeds, that's the same for all orgs, but orgs with status thrivable they needed to have planted 2000 Seeds by the time they became thrivable, so for instance in the case of a thrivable org which amount they can't unplant 200 or 2000 Seeds?

n13:
That is a good question - I guess we should allow to unplant this but then remove the status, but that's not good as it may accidentally happen. 

Ideally the org could easily move in and out of thrivable status, and unplanting should show the appropriate warnings on the UX - but that's not our problem in the smart contracts.

PR #343, opened 2021-06-03
Organization uses balance instead of planted score for tx limits #320

IanMendozaJaimes:
- Added validation in `harvest::unplant` so organizations can not unplant their fee
- Added tests

- [ ] Harvest

No migrations nor settings update are required

### DIFF
--- a/src/seeds.harvest.cpp
+++ b/src/seeds.harvest.cpp
@@ -240,6 +240,11 @@ void harvest::unplant(name from, asset quantity) {
   auto bitr = balances.find(from.value);
   check(bitr->planted.amount >= quantity.amount, "can't unplant more than planted!");
 
+  auto oitr = organizations.find(from.value);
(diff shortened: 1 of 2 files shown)


n13 (changes requested):
Hmmm... I think there was some confusion about the word "planted"

An org has a 200 Seeds fee to get created, this is the planted balance that can't be unplanted.

In the original implementation, I did not actually plant this value, but rather kept it in a balance in the org contract, so that it doesn't have the issue of possibly getting unplanted. 

So not sure we need this, or we need it this way.

We could switch the whole system to this, then planted score of the org would probably work more smoothly, no special cases for transactions limit and so on - 

I like this... discuss tonight.

n13 (approved):

n13:
Approved changes after talking to Ian - this is the right way to do it

n13:
Deployed

Issue #348, opened 2021-06-09
Multisig Strategy


n13:
======== **FOR REVIEW**

## Permissions Structure

All contracts owner has threshold 1, permissions cg.seeds@owner, msig.seeds@eosio.code
All contracts active has threshold 1, permissions cg.seeds@active, msig.seeds@eosio.code

cg permission allows msig to check for required permissions
msig.seeds@eosio.code allows the msig contract to execute proposals against contract active/owner permissions

cg.seeds has permissions for 5 signing accounts with 3/5 signing, and also msig.seeds@eosio.code with weight 3, so msig.seeds can execute proposals that change the account permissions.

msig.seeds itself has the same permissions as all other contracts, so it can process proposals that set code on itself.

## Tools

To make contract deployment and change permissions easier, I am creating the following tools as JS scripts, in our seeds.js deploy script:
Before deployment:
- Place contract under the above permission

Post deployment:
- propose to place contract under key permission again (in case something goes wrong, we have to revert)
- propose deployment of contract code and abi
- propose to change guardians on cg.seeds - a new list of guardians

n13:
Rieki likes shield.seeds

I also like guards.seeds even though it is so close to the light wallet guardians contract.

n13:
@dappdever @IanMendozaJaimes @7flash @sebastianmontero

PR #349, opened 2021-06-09
Multisig Tools

n13:
Tools / scripts to make msig proposals for contract deployment and permissions changes

See #348 

### proposeDeploy
Propose contract deployment, print ESR links and QR codes for approve and exec

### proposeChangeGuardians
Propose a different set of guardians, print ESR links and QR codes for approve and exec

### setCGPermissions
Change an account over to msig.seeds + cg.seeds permissions (multisig) - requires the key permissions for the old system

### proposeKeyPermissions
Revert contract back to key permissions - in case we need this

## Tasks 

- [x] proposeDeploy
- [x] proposeChangeGuardians
- [x] setCGPermissions (almost done, needs small changes)
- [x] proposeKeyPermissions

### DIFF
--- a/docs/accounts.html
+++ b/docs/accounts.html
@@ -30,25 +30,37 @@ For most account interactions, including moving from visitor to citizen, vouchin
 </comment></p>
 <h2>Actions</h2>
 <ul>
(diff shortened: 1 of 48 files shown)


n13:
tested on testnet and mainnet - these scripts won't work in all cases, but will work in many cases

PR #352, opened 2021-06-10
Fix referendum vote allication

n13:
added referendum scope to votes

referendums grabs voice from proposals

missing: need to go through and see if anyone has been removed from voice in the meantime, and remove them from referendum voice too. but - they won't get updated so in any case voice decay should zero them out over time once that's implemented.


n13:
referendum tests fail but they're so obscure they're hard to fix - the code works.

PR #353, opened 2021-06-10
rename mintts to minthrvst

n13:
Rename minttst to minthrvst since we will use it to mint

in the future adding permissions will be more complicated

## Deployment

remove existing minthrvst permission before deploying this, to be safe

```
  target: `${accounts.token.account}@minthrvst`,
  action: 'minttst'
  action: 'minthrvst'

```


IanMendozaJaimes (approved):
It looks great!

n13:
deployed

PR #354, opened 2021-06-12
Fixing Gratitude Tokens recursive new round

juliolrmonteiro:
This fixes the code for the `newround` action, that was failing on mainnet because it was running `calc_acks` over too many accounts at once, now we call `calc_acks` recursively.

Also took the opportunity to add acks stats and a new setting to determine how much % of the gratitude pot is saved for the next round (default 0, to behave as now).

### Instructions
- [x] Deploy: gratitude / settings
- [x] New settings / run updateSettings
- [x] Required migration: `gratitude::migratestats`
- [ ] No new Permissions

### DIFF
--- a/include/seeds.gratitude.hpp
+++ b/include/seeds.gratitude.hpp
@@ -24,6 +24,7 @@ CONTRACT gratitude : public contract {
         balances(receiver, receiver.value),
         acks(receiver, receiver.value),
         stats(receiver, receiver.value),
+        stats2(receiver, receiver.value),
         sizes(receiver, receiver.value),
         users(contracts::accounts, contracts::accounts.value),
         config(contracts::settings, contracts::settings.value)
@@ -41,7 +42,7 @@ CONTRACT gratitude : public contract {
     ACTION newround();
 
     // Called after all acks are calculated
-    ACTION payround();
+    ACTION payround(uint64_t start, uint64_t usable_bal);
 
     // Recursivelly calculate acks
     ACTION calcacks(uint64_t start);
@@ -108,6 +109,14 @@ CONTRACT gratitude : public contract {
     TABLE stats_table {
       uint64_t round_id;
       uint64_t num_transfers;
+      asset volume;
+
+      uint64_t primary_key() const { return round_id; }
+    };
+
+    TABLE stats_table_v2 {
+      uint64_t round_id;
+      uint64_t num_transfers;
       uint64_t num_acks;
       asset round_pot;
       asset volume;
@@ -124,9 +133,12 @@ CONTRACT gratitude : public contract {
 
     typedef eosio::multi_index<"stats"_n, stats_table> stats_tables;
 
+    typedef eosio::multi_index<"stats2"_n, stats_table_v2> stats_tables_v2;
+
     balance_tables balances;
     acks_tables acks;
     stats_tables stats;
+    stats_tables_v2 stats2;
 
     // External tables
     user_tables users;
--- a/include/seeds.gratitude.hpp
+++ b/include/seeds.gratitude.hpp
@@ -61,6 +61,7 @@ CONTRACT gratitude : public contract {
 
     void check_user(name account);
     void init_balances(name account);
+    void reset_balances(name account);
     void _calc_acks(name account);
     void add_gratitude(name account, asset quantity);
     void sub_gratitude(name account, asset quantity);
--- a/include/seeds.gratitude.hpp
+++ b/include/seeds.gratitude.hpp
@@ -40,13 +40,25 @@ CONTRACT gratitude : public contract {
     // Generates a new gratitude round, regenerating gratitude and splitting stored SEEDS
     ACTION newround();
 
+    // Called after all acks are calculated
+    ACTION payround();
+
+    //  Recursivelly calculate acks
+    ACTION calcacks(uint64_t start);
+
+    //  Calculate acks for testing
     ACTION testacks();
+
+
+    // Called when depositing SEEDS into the pot
+    ACTION deposit(name from, name to, asset quantity, string memo);
+
 
   private:
 
     void check_user(name account);
     void init_balances(name account);
-    void calc_acks(name account);
+    void _calc_acks(name account);
     void add_gratitude(name account, asset quantity);
     void sub_gratitude(name account, asset quantity);
     uint64_t get_current_volume();
@@ -94,6 +106,8 @@ CONTRACT gratitude : public contract {
     TABLE stats_table {
       uint64_t round_id;
       uint64_t num_transfers;
+      uint64_t num_acks;
+      asset round_pot;
       asset volume;
 
       uint64_t primary_key() const { return round_id; }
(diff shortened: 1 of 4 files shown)


src/seeds.gratitude.cpp
```diff
@@ -28,11 +28,33 @@ ACTION gratitude::reset () {
...
     item.volume = asset(0, gratitude_symbol);
+    item.round_pot = asset(0, seeds_symbol);
   });
 
 }
 
+ACTION gratitude::migratestats() {
+  require_auth(get_self());
+
+  name old = "gratz.seeds"_n;
```
n13, line 41: Not sure how this migration is supposed to work?

did you change the table name?

src/seeds.gratitude.cpp
```diff
@@ -70,38 +92,88 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
...
-ACTION gratitude::newround() {
+ACTION gratitude::calcacks(uint64_t start) {
+  require_auth(get_self());
+
+  auto actr = start == 0 ? acks.begin() : acks.find(start);
+
+  _calc_acks(actr->donor);
+
+  actr++;
+
```
n13, line 128: Normally, we process a bunch of items at the same time, e.g. 40 or whatever, 1 at a time is slow?!

src/seeds.gratitude.cpp
```diff
@@ -70,38 +92,88 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
...
+      std::make_tuple(actr->donor.value)
+    );
+
+    transaction tx;
+    tx.actions.emplace_back(next_execution);
+    tx.delay_sec = 1;
+    tx.send(actr->donor.value, _self);
+  } else {
+    // Cleanup on last exec
+    auto actr2 = acks.begin();
```
n13, line 144: This will have the same problem as the initial issue, as it's walking the entire table to delete items...

This will fail if the number of acks is > 200 or 300 or 400  (one of these numbers)

n13: I would suggest to erase them as they're processed, if that is possible

src/seeds.gratitude.cpp
```diff
@@ -70,38 +92,88 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
...
-  while (actr != acks.end()) {
-    calc_acks(actr->donor);
-    actr = acks.erase(actr);
-  }
+  float potkeep = config_get(gratz_potkp) / (float)100;
+  uint64_t usable_amount = contract_balance.amount - (contract_balance.amount * potkeep);
 
   auto bitr = balances.begin();
   while (bitr != balances.end()) {
     uint64_t my_received = bitr->received.amount;
```
n13, line 172: This is also a candidate where if the number of balances > 200 or so, it will fail

So best to do everything you want to do on all the accounts in the recursive (not really recursive but OK) action

n13 (changes requested):
Request changes so we can be sure that nothing times out if this scales to many users

src/seeds.gratitude.cpp
n13: I see payround maybe a second batch action, after all acks are done

src/seeds.gratitude.cpp
juliolrmonteiro: In this case, each account can have around 10 transfers, so I was thinking of doing one account at a time, but we may batch 10 to test.

src/seeds.gratitude.cpp
juliolrmonteiro: Fixed

src/seeds.gratitude.cpp
n13: oh thats ok then

src/seeds.gratitude.cpp
```diff
@@ -70,61 +99,173 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
...
+  auto actr = start == 0 ? acks.begin() : acks.find(start);
 
-  auto actr = acks.begin();
-  while (actr != acks.end()) {
-    calc_acks(actr->donor);
+  if (actr != acks.end()) {
+    _calc_acks(actr->donor);
     actr = acks.erase(actr);
   }
+  // If there is still more, do recursion call
```
n13, line 136: did you test that the recursive calling works? for example, has the permission, eosio.code on active, etc, 

I usually give it a batchsize parameter so that I can test it with batch size 1 in the unit tests

But if you tested it, cool, just checking and making sure.

src/seeds.gratitude.cpp
```diff
@@ -70,61 +99,173 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
         item.receivers.push_back(to);
     });
   }
+
+  // Updates ack stats
+  auto stitr = stats2.rbegin();
+  auto round_id = stitr->round_id;
+
+  auto stitr2 = stats2.find(round_id);
```
n13, line 107: you don't need sitr2 here at all, stitr == stitr2 ...

1 you are getting the round_id (the primary key) from stitr
2 - you're searching for round_id to get stitr2 -> stitr2 will always be == stitr

src/seeds.gratitude.cpp
```diff
@@ -70,61 +99,173 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
...
+    auto round_id = stitr->round_id;
+
+    auto stitr2 = stats2.find(round_id);
+    auto oldpot = stitr2->round_pot.amount;
+    stats2.modify(stitr2, _self, [&](auto& item) {
+        item.round_pot = asset(oldpot + quantity.amount, seeds_symbol);
+    });
+
+    // // Continue with transfer
+    // token::transfer_action action{contracts::token, {_self, "active"_n}};
```
n13, line 254: remove commented code?

src/seeds.gratitude.cpp
```diff
@@ -70,61 +99,173 @@ ACTION gratitude::acknowledge (name from, name to, string memo) {
...
+      to  ==  get_self() &&                     // to here
+      quantity.symbol == seeds_symbol) {        // SEEDS symbol
+
+    utils::check_asset(quantity);
+
+    // Updates stats
+    auto stitr = stats2.rbegin();
+    auto round_id = stitr->round_id;
+
+    auto stitr2 = stats2.find(round_id);
```
n13, line 247: same here, stitr2 not needed?!

include/seeds.gratitude.hpp
```diff
@@ -120,8 +149,25 @@ CONTRACT gratitude : public contract {
     const name gratzgen_res = "gratz1.gen"_n; // Gratitude generated per cycle for residents
     const name gratzgen_cit = "gratz2.gen"_n; // Gratitude generated per cycle for citizens
     const name gratz_acks = "gratz.acks"_n; // Gratitude generated per cycle for citizens
+    const name gratz_potkp = "gratz.potkp"_n; // Percentil of the pot to keep each round
```
n13, line 152: if this is configurable, there needs to be a unit test that tests for this being != 0

src/seeds.settings.cpp
```diff
@@ -331,6 +331,7 @@ void settings::reset() {
   confwithdesc(name("gratz1.gen"), 100 * 10000, "Base quantity of gratitude tokens received for residents per cycle", medium_impact);
   confwithdesc(name("gratz2.gen"), 200 * 10000, "Base quantity of gratitude tokens received for citizens per cycle", medium_impact);
   confwithdesc(name("gratz.acks"), 10, "Minumum number of gratitude acks to fully use your tokens", medium_impact);
+  confwithdesc(name("gratz.potkp"), 0, "Percent to keep on Gratitude Pot at new round", medium_impact);
```
n13, line 334: if this is configurable, there needs to be a unit test that tests for this being != 0

n13: So this is my change request, the other things are all just comments

n13 (changes requested):
confwithdesc(name("gratz.potkp"), 0, "Percent to keep on Gratitude Pot at new round", medium_impact);


^^ this needs a unit test

src/seeds.gratitude.cpp
juliolrmonteiro: Done.

src/seeds.gratitude.cpp
juliolrmonteiro: Yes, tested.

src/seeds.gratitude.cpp
juliolrmonteiro: Oh, sure!

include/seeds.gratitude.hpp
juliolrmonteiro: Added the test, thanks!

src/seeds.settings.cpp
juliolrmonteiro: Added the new test, as requested.
The changes from stitr2 where not possible because `rbegin()` returns a `const` iterator which doesn't allow call to `modify()`, I tried that before. :)

n13 (approved):

PR #355, opened 2021-06-13
Moon op improvements

n13:
added some code to make it easier to schedule moon ops

added some safety check so it's impossible to add a past op (as that would trigger an immediate firing, which would be bad if we have "incprice" or "run proposals"... ;) whoops!!!

Added "addmoonop" so we don't have to figure out unix time stamps - just set the phase as string, and it figures it out.

I also added one to testnet let's see if it fires next time, then can move to mainnet
'''
cleos -u "https://test.hypha.earth" push action cycle.seeds addmoonop '{ 
        "id":"inc.lprice",
        "action":"incprice",
        "contract":"tlosto.seeds",
        "quarter_moon_cycles":"1",
        "start_phase_name":"First Quarter"
}' -p cycle.seeds@active

> cleost get table cycle.seeds cycle.seeds moonops
{
  "rows": [{
      "id": "inc.lprice",
      "action": "incprice",
      "contract": "tlosto.seeds",
      "quarter_moon_cycles": 1,
      "start_time": 1623988440,
      "last_moon_cycle_id": 0,
      "pause": 0
    }
  ],
  "more": false,
  "next_key": ""
}

'''

### DIFF
--- a/include/seeds.scheduler.hpp
+++ b/include/seeds.scheduler.hpp
@@ -52,6 +52,7 @@ CONTRACT scheduler : public contract {
         void exec_op(name id, name contract, name action);
         void cancel_exec();
         void reset_aux(bool destructive);
+        uint64_t next_valid_moon_phase(uint64_t moon_cycle_id, uint64_t quarter_moon_cycles);
         bool should_preserve_op(name op_id) {
             return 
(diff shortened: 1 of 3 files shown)


IanMendozaJaimes (approved):
Looks great

PR #357, opened 2021-06-14
Onboarding: Don't accept existing account in accept action

n13:
## Hotfix - Don't accept existing Telos accounts

Reason / user story: One of our users signed on with an existing Telos account - not sure how that was possible in the app, but it was possible.

In reality this means she signed on with an account that was not her account! 

She did not realize this until much later, when 15,000 Seeds had been deposited into her account and she tried to use them.

The passport app had stored the public-private key pair that was created at account creation but that keypair was never actually used, because no new account was created.

Pic from the support channel chat

image

## Other changes
Also: 
minor bugfixes - campinvite auth check didn't work with the random account, replaced binary search with code I can understand and it worked ;) 
fix unit tests


IanMendozaJaimes (approved):
Looks great!

PR #358, opened 2021-06-15
Feature/rebuild referendums

IanMendozaJaimes:
The new Referendums structure. 

All the referendums now will inherit from the class Referendum and each class will have its own implementation of the basic referendum operations: `create`, `update`, `cancel` and `evaluate`.

I also added a field in the `referendums_table` which is `std::map<string, VariantValue> special_attributes`. The idea is that we could have an object like in javascript that can hold several types and we can store the special attributes related to each referendum type, for instance a referendum that wants to change a setting will have different special_attributes than a referendum that wants to push new code.

## Deployment

Contracts:
- [ ] Referendums

Other considerations:
- [ ] Settings need to be updated

Migrations might be required


n13:
Ok what types do we have?

I think
settings (changes settings, float or int)
constitution / game guide update (changes game guide)
generic (changes nothing on chain, like the one we have at the moment)
code (propose a code change, this one would provide the new code and abi in tables, like we do with multisig, it would execute a msig proposal basically, which would call set_code and set_abi - no need to have this for the 1st release, just mentioning

IanMendozaJaimes:
- Added referendums refactor
- Moved voice methods into referendums contract (not all of them and not all of them are being called)
- Rebuild tests

Contracts:
- [ ] Referendums

Other considerations:
- [ ] Settings need to be updated

Migrations might be required

include/proposals/referendum_settings.hpp
```diff
@@ -0,0 +1,52 @@
...
+    } 
+    SettingInfo;
+
+  private:
+
+    static constexpr name high_impact = name("high");
+    static constexpr name medium_impact = name("med");
+    static constexpr name low_impact = name("low");
+
+    uint64_t get_required_unity(const name & setting, const bool & is_float);
```
n13, line 40: shouldn't this return a float?

include/tables/proposals_table.hpp
```diff
@@ -0,0 +1,55 @@
...
+      string url; \
+      time_point created_at; \
+      name status; \
+      name stage; \
+      name type; \
+      uint64_t last_ran_cycle; \
+      uint64_t age; \
+      EOSLIB_SERIALIZE(proposal_table, (proposal_id)(favour)(against)(staked)(creator)(title) \
+        (summary)(description)(image)(url)(created_at)(status)(stage)(type)(last_ran_cycle)(age)) \
+      \
```
n13, line 28: never seen this statement before, what does it do?

scripts/eosjs-port.js
```diff
@@ -104,19 +104,27 @@ class Eos {
...
-          }
-        } else {
-          if ((action.fields.length + 1) != arguments.length) {
-            throw new Exception(`Not enough arguments to call ${action.name} action in ${accountName} contract`)
-          }
-          for (let i = 0; i < action.fields.length; i++) {
-            const { name } = action.fields[i]
-            data[name] = arguments[i]
-          }
+        // if (action.fields.length > 0 && arguments.length == 2 && typeof arguments[0] === 'object') {
```
n13, line 107: What about this commented out code? Replaced by new simpler code?

scripts/helper.js
```diff
@@ -711,8 +711,10 @@ var permissions = [{
 }, {
   target: `${accounts.history.account}@execute`,
   action: 'cleanptrxs'
-}
-]
+}, {
```
n13, line 714: This is the same as previously proposals right, we can add to reputation... maybe this permission can be limited to addrep. 

Although I guess in the end all contracts will be under the same permissions anyway...

src/proposals/referendums_settings.cpp
```diff
@@ -0,0 +1,393 @@
...
+  check(ritr->stage == ProposalsCommon::stage_staged, "can not cancel referendum, it is not staged");
+  check(ritr->type == ProposalsCommon::type_ref_setting, "referendum has to be of type settings");
+
+  proposals_t.erase(ritr);
+  propaux_t.erase(raitr);
+
+}
+
+void ReferendumSettings::evaluate (std::map<std::string, VariantValue> & args) {
+
```
n13, line 126: what is the reason to pass an args object in here, vs. just the id?

n13:
All looks good, but huge amount of code, I will go over this n detail once the unit test is working

Only problem was, unit test not passing

```
./scripts/seeds.js test referendums
Local
-  
    Local
  Referendums Settings

    reset referendums
    reset settings
    reset accounts
    init propcycle
    insert test settings
    join users
    create referendum
    update referendum
    delete referendum    ✖  ReferenceError: table is not defined
    ----------------------------------------
        operator: deepEqual
        diff: 
        source: at runMicrotasks (<anonymous>)
        at processTicksAndRejections (internal/process/task_queues.js:93:5)
        
        



  passed: 0,  failed: 1  of 1 tests  (703ms)


```

n13:
The conflicts... can probably be ignored because that's all the old codebase, no need to deal with that

include/proposals/referendum_settings.hpp
IanMendozaJaimes: Well, in the case of referendums they were getting the unity and the quorum needed looking at the `config` table to get the values depending on the impact. It is an integer because the `config` table stores integers. Do you think we should move those settings to the `floatconfig` table?

include/tables/proposals_table.hpp
IanMendozaJaimes: That's an statement that tells the compiler that those are the attributes that needs to serialize to get the abi correctly generated. In that table it's needed I can remove it, but it is needed in the referendums_auxiliary table so the variant type is properly generated

scripts/eosjs-port.js
IanMendozaJaimes: Yes, as I had to pass an array now for calling the functions I updated the way it was handling the parameters.

scripts/helper.js
IanMendozaJaimes: ok, I will fix it

src/proposals/referendums_settings.cpp
IanMendozaJaimes: There is no difference for the moment, but I in the proposals contract for the evaluation we pass the cycle as well and as each proposal will implement it's own evaluate method I didn't want to limit the Parent class, but I can change it

IanMendozaJaimes:
I will check the tests, maybe something was broken when I merged

include/proposals/referendum_settings.hpp
n13: No it's fine

n13 (approved):

n13:
all teste passing, merging

n13:
I was merging some code in eos-port can you check this is now OK @IanMendozaJaimes

PR #359, opened 2021-06-16
added remove method for cleanup purposes

n13:
just added a remove action because there's some stray settings on mainnet we need to clean up


IanMendozaJaimes (approved):
Looks good to me

PR #364, opened 2021-06-16
Better test for check cleanup onboarding action

EdwinFajardoB:
This should 
- reference the issue it's based on
- describe briefly what this is all about - why this change was made
- describe which contracts need to be deployed, settings update, scheduler update, permissions update (Y/N on all these)

### DIFF
--- a/src/seeds.onboarding.cpp
+++ b/src/seeds.onboarding.cpp
@@ -408,12 +408,7 @@ void onboarding::accept(name account, checksum256 invite_secret, string publicKe
 void onboarding::chkcleanup() {
   require_auth(get_self());
 
-  action(
-      permission_level{get_self(), "active"_n},
-      get_self(),
-      "chkcleanupau"_n,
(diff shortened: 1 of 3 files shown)


IanMendozaJaimes:
Contracts to be modified:
- [ ] Onboarding

src/seeds.onboarding.cpp
```diff
@@ -408,6 +408,17 @@ void onboarding::accept(name account, checksum256 invite_secret, string publicKe
 void onboarding::chkcleanup() {
   require_auth(get_self());
 
+  action(
+      permission_level{get_self(), "active"_n},
+      get_self(),
+      "chkcleanupau"_n,
+      std::make_tuple((utils::seconds_per_day * 30))
+  ).send();
+}
```
n13, line 417: I don't understand what is the benefit of this?

n13 (changes requested):
Does this pertain to any sort of issue or bug? I don't know what the point is of this change.

PR #370, opened 2021-06-30
revert votes does not use a user's votes

### LINKED ISSUE #369: Add reverse trust action in proposals and referendums
The current reverse trust uses the "against" action - I think that has some bugs / unintended consequences. 

- it's sensitive to the number in "against" - it should reverse the originally cast number, so if it was favour 5, it should now be against 5 (same amount)
- It's using up the user's votes - example if they voted 99 points for the proposal, they now lose 99 votes from their current cycle

##Spec

- Add reversetrust(user, prop_id) action
- Remove reverse trust code from "against" - this should make the code cleaner as well

### Reverse trust
- Changes the user's "favour" vote to an "against" vote on a proposal in eval stage
- If they did not vote favour, error
- if proposal not in eval state, error
- Does not use up any votes the user has for the current cycle

n13:
also implemented for delegation

## What

Revert votes needs to just revert a user's votes, not use votes from the current vote pool - so on an eval proposal, a revert does not consume votes. It just reverts a yes to a no with the same number of votes.

Fixes #369 for proposals.

## Note
Was a bit tricky with vote delegation but I modified the unit tests for reversal of notes so have some confidence it works.

## Contracts

proposals

No other contracts affected

### DIFF
--- a/include/seeds.proposals.hpp
+++ b/include/seeds.proposals.hpp
@@ -61,6 +61,9 @@ CONTRACT proposals : public contract {
 
       ACTION neutral(name user, uint64_t id);
 
+      ACTION revertvote(name user, uint64_t id);
+      ACTION mimicrevert(name delegatee, uint64_t delegator, name scope, uint64_t proposal_id, uint64_t chunksize);
+
       ACTION voteonbehalf(name voter, uint64_t id, uint64_t amount, name option);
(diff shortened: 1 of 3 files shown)


IanMendozaJaimes (approved):
It looks great to me

PR #383, opened 2021-07-20
check if the user balance has been inited when calculating acks

juliolrmonteiro:
Checks if the user balance has been inited when calculating acks. This bug should only happen when mixing acks being created with the old version of the gratitude contract, that didn't init that account.


n13 (approved):
looks good

PR #393, opened 2021-09-04
Feature/streaming funding

IanMendozaJaimes:
Includes DAO https://github.com/JoinSEEDS/seeds-smart-contracts/pull/390

Added streaming funding.

Added three new tables:
- `dho_table` : for storing the organizations that are dhos
- `dho_vote_table` : for storing user's votes, it is scoped by contract's scope
- `dho_shares_table` : for storing the dhos that will receive part of the harvest 

Added tests for dao, harvest and scheduler.

**Brief description**

Dhos can be created and deleted with `createdho` and `removedho` respectively.
When a dho is removed, all its votes are removed as well and the dho is removed immediately from the sharing table.

To vote, users must call the action `votedhos` with all the dhos they want to vote for. I think this way is easier to implement and validate. Users can call this action at any time and the votes will be recalculated for that user except if the user's voice is delegated. The delegation is implemented as well, to delegate it the dao actions delegate/undelegate are used.

To calculate how much percentage of the harvest each dho receives, the function `dhocalcdists` has to be triggered. I set this function to be executed every hour in scheduler.

To clean the old votes, the function `dhocleanvts` has to be called. I set this function to be executed every day.

**Some considerations:**
- I added the dho_voice scope so the voice delegation can use the same methods for delegate/undelegate
- For the cs points multiplier I am currently using the rank, i.e. from 1-99, I decided that because with the normal 0-2 multiplier votes percentages can become 0 very easily.

## Deployment

Contracts to be deployed:
- [ ] Dao
- [ ] Harvest
- [ ] scheduler

Other considerations:
- [ ] Settings need to be updated


include/proposals/proposal_campaign_funding.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "proposals_base.hpp"
+
```
n13, line 4: FWIW these campaigns will likely no longer exist going forward - except invite funding campaigns.

include/proposals/proposals_base.hpp
```diff
@@ -5,6 +5,10 @@
 namespace ProposalsCommon {
   constexpr name type_ref_setting = name("r.setting");
   constexpr name type_ref_code = name("r.code");
+  constexpr name type_prop_alliance = name("p.alliance");
```
n13, line 8: How about adding referendums as another proposal type here?

I think that was the idea, to combine the two

include/proposals/proposals_factory.hpp
```diff
@@ -12,6 +16,18 @@ class ProposalsFactory {
       {
       case ProposalsCommon::type_ref_setting.value:
         return new ReferendumSettings(_contract);
```
n13, line 18: Ok got it

n13:
Maybe tonight we can go over this code after the meeting (I mean your morning) so I can make sure I understand it all

I will go through a full proposal cycle reading the code first as well, so I am prepared.

include/proposals/proposals_base.hpp
```diff
@@ -5,6 +5,10 @@
 namespace ProposalsCommon {
   constexpr name type_ref_setting = name("r.setting");
```
n13, line 6: lets call this ref.setting, and ref.code ;)

include/proposals/proposals_base.hpp
```diff
@@ -30,23 +34,38 @@ class Proposal {
...
-    virtual void cancel(std::map<std::string, VariantValue> & args) = 0;
+    virtual name get_scope() = 0;
+    virtual name get_fund_type() = 0;
 
-    virtual void evaluate(std::map<std::string, VariantValue> & args) = 0;
 
-    virtual name get_scope() = 0;
+    virtual void check_can_vote(const name & status, const name & stage);
+    virtual bool check_prop_majority(std::map<std::string, VariantValue> & args);
+    virtual uint64_t min_stake(const asset & quantity, const name & fund);
```
n13, line 54: Would be nice to keep everything similar in the API

we usually use this func(name x, name y)

Why is this func(const name & x, const name & y) ?

I can see how const makes sense, as it leads to cleaner code, but the &... is this needed? Does it change anything?

n13: (not a c++ expert here ;) )

include/seeds.dao.hpp
```diff
@@ -31,17 +33,42 @@ CONTRACT dao : public contract {
...
+      name campaign_fund = "campaign"_n;
+      name milestone_fund = "milestone"_n;
+
+      std::vector<name> fund_types = {
+        alliance_fund,
+        campaign_fund,
+        milestone_fund
+      };
+
+      const name prop_active_size = "prop.act.sz"_n;
```
n13, line 59: Are we actually keeping track of all these sizes, or are they carryover from the other contracts?

I know that we had them in the old contracts, but not sure they are still all in use.

include/seeds.dao.hpp
```diff
@@ -31,17 +33,42 @@ CONTRACT dao : public contract {
...
+      const name user_active_size = "user.act.sz"_n;
+      const name cycle_vote_power_size = "votepow.sz"_n;
+      const name dhos_vote_size = "dho.vote.sz"_n; 
+      const name linear_payout = "linear"_n;
+      const name stepped_payout = "step"_n;
+
+      typedef struct dhovote {
+        name dho;
+        uint64_t points;
```
n13, line 68: points here is always 0-100 right, like percentage points...

include/proposals/proposals_base.hpp
IanMendozaJaimes: Well, if all the functions in the contract use &, it can lead to a better performance, because now we're only using references and a reference only uses 8 bytes (usually), so in terms of memory passed among all the functions in the contract it's really small. In that way we can pass types that are heavier than 8 bytes like names or maps but we're only consuming 8 bytes.

The reference is probably useless in the action declaration like `ACTION dao::create()`, but it's definitely needed in the subsequent functions called inside `create`.

include/seeds.dao.hpp
IanMendozaJaimes: yes, the dao contract uses all those sizes. I am not sure all of them are in use, I think the `votepow.sz`is not used but I included anyways just to try to mimic proposals behavior. There are four sizes but only the votepow.sz is the one that I think is not needed

include/seeds.dao.hpp
IanMendozaJaimes: amm well no, I am storing there the result of multiplying:
`assigned percentage * user's contribution score`
so it can be greater than 100. 

When a user votes, the user needs to call `votedhos`, in that function the contract validates that the percentage the user is assigning sums up to 100. The user only has to vote for the orgs he is going to assign a percentage bigger than 0, currently the minimum is 1% because it checks the sum is 100, I would have to modify it to sum up to 10000 if you would like to include 2 decimals.

include/proposals/proposals_base.hpp
IanMendozaJaimes: ok, I will change it

include/seeds.dao.hpp
n13: I don't think we need decimals, and thanks for the explanation

include/proposals/proposals_base.hpp
n13: Well unlike desktop systems, in smart contracts this kind of thing can actually make a difference, thanks for the explanation, let's leave it.

n13 (approved):

PR #403, opened 2021-10-01
 dao.seeds allow voice scopes being added or removed #395 

### LINKED ISSUE #395: dao.seeds needs to easily allow voice scopes being added or removed
## Use case
- We add a new category of proposal called "cats"
- users receive voice scopes for cats
- voting happens on that scope

## Use case 2
- we remove category cats again - that prop type no longer exists
- voice scopes for cats are removed

Just want to make sure that this is easily possible.

tlacloc:
Fix #395

Actions added 
- deletescope
- addvoice

Actions updated
- updatevoice
- updatevoices

Changes made:
`updatevoice` now updates all voices or just a one.
`updatevoices` small change because it calls `updatevoice`.
`deletescope` deletes all voice values from all users for a given proposal scope contained in `scopes` vector.
`addvoice` add all voice values from all users, with its respective value (decay calculated) for a given proposal scope contained in `scopes` vector.

**Contracts to be deployed:**
- [ ] dao.seeds

Other considerations:
- No settings added
- No permissions added
- No migrations required

### DIFF
--- a/src/proposals/voice_management.cpp
+++ b/src/proposals/voice_management.cpp
@@ -3,10 +3,10 @@
 
 ACTION dao::updatevoices () {
   require_auth(get_self());
-  updatevoice(uint64_t(0));
+  updatevoice(uint64_t(0), ""_n);
 }
 
-ACTION dao::updatevoice (const uint64_t & start) {
+ACTION dao::updatevoice (const uint64_t & start, const name & scope) {
   require_auth(get_self());
   
   DEFINE_CS_POINTS_TABLE
@@ -30,8 +30,8 @@ ACTION dao::updatevoice (const uint64_t & start) {
       if (csitr != cspoints_t.end()) {
         points = csitr->rank;
       }
-      print("account: ", vitr->account, ", points: ", points, "\n");
-      set_voice(vitr->account, points, ""_n);
+      print("account: ", vitr->account, ", points: ", points, scope);
+      set_voice(vitr->account, points, scope);
       if (is_active(vitr -> account, cutoff_date)) {
         vote_power += points;
         active_users++;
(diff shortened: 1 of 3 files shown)


src/proposals/voice_management.cpp
```diff
@@ -46,11 +46,95 @@ ACTION dao::updatevoice (const uint64_t & start) {
...
+
+  for (auto & s : scopes) {
+    if (s == scope) { existing_scope = true; }
+  }
+
+  check(existing_scope, "scope must exist");
+
+  cs_points_tables cspoints_t(contracts::harvest, contracts::harvest.value);
+
+  voice_tables voices_t(get_self(), campaign_scope.value);
```
n13, line 104: Ok so the campaign voice is used here as a list of "citizens who have voice" right?

Then we just use the account names and call setvoice on them....

src/proposals/voice_management.cpp
```diff
@@ -46,11 +46,95 @@ ACTION dao::updatevoice (const uint64_t & start) {
...
+      voice_amount = calculate_decay(csitr->rank);
+    }
+
+    print("account: ", vitr->account, ", points: ", voice_amount, scope);
+    set_voice(vitr->account, voice_amount, scope);
+
+    vitr++;
+    count++;
+  }
+
```
n13, line 124: Now that we added a new voice scope - how can the contract know which scopes exist?

We don't store the scopes anywhere, do we?

lets say I add a new voice scope called "foobar"

Then that voice scope won't get decayed, right?

n13 (changes requested):
Not sure about how new voice scopes will be stored, where they are stored, and who can handle them. 

I can think of decay which needs to decay all scopes, but there may be other code that needs to iterate through all scopes.

src/proposals/voice_management.cpp
tlacloc: Well, there is a vector of scopes, also addvoices check for scopes in that vector

n13 (approved):

PR #408, opened 2021-11-03
Fix stake burning issue

### LINKED ISSUE #401: Bugfix: Return stake
Before: Lose 100% if proposal fails for any reason

Desired / After:
Lose 5% if it doesn't reach quorum
Lose 100% if it doesn't reach unity

## Spec
If unity is not met, burn the stake (same as before)
if unity is met, but quorum is not met, burn part of the stake, return the rest to the sponsor (whoever paid for it to begin with)
if both unity and quorum are met, the whole stake is returned (as before)

tlacloc:
Bugfix: Return stake
 Fix #401 
 
Now when a proposal is rejected (passing unity but with invalid quorum) user gets 95% refund

Changes in 

- [ ] seeds.proposals
- [ ] seeds.dao


src/seeds.proposals.cpp
```diff
@@ -528,8 +528,13 @@ void proposals::evalproposal (uint64_t proposal_id, uint64_t prop_cycle) {
       }
 
     } else {
+
+      asset stackToBurn = valid_quorum ?  pitr->staked : asset(pitr->staked.amount * 0.05, seeds_symbol);
```
n13, line 532: typo - should be "stakeToBurn" - please fix

n13: Code looks great.

src/seeds.proposals.cpp
```diff
@@ -528,8 +528,13 @@ void proposals::evalproposal (uint64_t proposal_id, uint64_t prop_cycle) {
...
+      stackToBurn = passed ? stackToBurn : pitr->staked;
```
n13, line 533: This is a little weird - an explicit if statement that defines all possible cases would be easier to understand. 

But - it's correct code.

src/seeds.proposals.cpp
```diff
@@ -547,7 +552,7 @@ void proposals::evalproposal (uint64_t proposal_id, uint64_t prop_cycle) {
             proposal.passed_cycle = prop_cycle;
           }
           proposal.executed = false;
-          proposal.staked = asset(0, seeds_symbol);
+          proposal.staked = asset(pitr->staked.amount - stackToBurn.amount, seeds_symbol);
           proposal.status = status_rejected;
```
n13, line 556: Ok but it should also send back the stake now. Or does this statement alone allow the user to claim back their stake?

n13 (changes requested):
Only question is this supposed to automatically send back the partial stake when unity is met, but quorum is not met?

src/seeds.proposals.cpp
n13: I updated the ticket to make this much clearer. Stake should be returned.

src/seeds.proposals.cpp
tlacloc: Oh, my bad, fixing it!

src/seeds.proposals.cpp
tlacloc: Thanks i wasn't that sure if it was suppose to store de value of the remained or should be returned, I fix this in the new commit too

n13 (approved):

PR #410, opened 2021-11-04
Added flag delegation

### LINKED ISSUE #406: Delegate Flagging
Just like voting is delegated we want to delegate flagging as well

IanMendozaJaimes:
Fixes #406 

- Added flag delegation
- Added tables `delegators` in accounts

#### Note
I added two generic functions in`utils`:
- Send deferred transaction
- Delete a table (for use it in the reset function)

I only added those functions to the actions involved in this issue. I think it could help make the contracts look cleaner. Please let me know what you think, I could refactor some contracts to try or if you feel it's not needed I can remove it and change it for the normal way.

#### Modified contracts:
- [ ] Accounts

#### Other considerations
No permissions nor settings need to be updated

### DIFF
--- a/include/utils.hpp
+++ b/include/utils.hpp
@@ -1,11 +1,13 @@
 #include <eosio/eosio.hpp>
 #include <eosio/asset.hpp>
 #include <eosio/system.hpp>
+#include <eosio/transaction.hpp>
 #include <tables/rep_table.hpp>
 #include <tables/size_table.hpp>
 #include <tables/user_table.hpp>
 #include <tables/config_table.hpp>
 #include <tables/config_float_table.hpp>
+#include <tables/deferred_id_table.hpp>
 
 using namespace eosio;
 using std::string;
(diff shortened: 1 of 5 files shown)


include/utils.hpp
```diff
@@ -226,6 +228,51 @@ namespace utils {
     return date.utc_seconds;
   }
```
n13, line 230: I love the util functions, really great!

include/utils.hpp
```diff
@@ -226,6 +228,51 @@ namespace utils {
...
+
+  template <typename... T>
+  void send_deferred_transaction (
+    const name & code,
+    const permission_level & permission,
+    const name & contract, 
+    const name & action,  
+    const std::tuple<T...> & data) {
+
+    DEFINE_DEFERRED_ID_TABLE
```
n13, line 251: I understand how these are supposed to work, creating a new unique deferred ID for each action

But I wonder and maybe you know this - what's the scope of a deferred ID? Is it a contract scope? Or global EOSIO network scope? Or signer scope? Payer scope? 

Example if I send a deferred transaction from accts.seeds with id = 1
And some other contract sends a deferred action to accts.seeds from another contract also with id 1 - is that the same ID?

I looked that up at some point but maybe you know what these IDs are unique to.

src/seeds.accounts.cpp
```diff
@@ -85,6 +85,7 @@ void accounts::reset() {
     banitr = ban.erase(banitr);
   }
 
+  utils::delete_table<delegators_tables>(contracts::accounts, contracts::accounts.value);
```
n13, line 89: cool! def the way to go forward.

src/seeds.accounts.cpp
```diff
@@ -1521,7 +1530,7 @@ void accounts::evaldemote (name to, uint64_t start_val, uint64_t chunk, uint64_t
     transaction tx;
     tx.actions.emplace_back(next_execution);
     tx.delay_sec = 1;
-    tx.send(next_value + 1, _self); 
+    tx.send(next_value + 1, _self);
   }
```
n13, line 1534: I think this also needs to call mimic, with remove?

n13 (changes requested):
removeflag also needs to be delegated...

include/utils.hpp
IanMendozaJaimes: The scope of the deferred id is the ram payer, in seeds it happens to be always the contract. So in this case the ids would be `accts.seeds::1`, and the other contract `other::1`, so there is no conflict. 

There is no conflict in using the id=1 several times, the only condition is that a deferred transaction shouldn't be scheduled with that id. What I mean is that, if the transaction with id 1 is already executed, then that id is free again and can be reused.

I added that method to not worry about deferred ids, because I see that in the contracts the id is always the next.value or something like that and there is some repeated code to create the transaction. And if at some point deferred transactions are removed from eosio, we could change that single method to use some scheduler and automatically fix all the deferred code.

src/seeds.accounts.cpp
IanMendozaJaimes: That's true, sorry I am going to fix it.

n13 (approved):

PR #411, opened 2021-11-07
refundstake added

n13:
App had many errors causing users to send too much stake to the contract

Adding refundstake action so they can get their Seeds back


IanMendozaJaimes (approved):
I looks good to me.

Issue #413, opened 2021-11-20
Vote Delegation Fails


n13:
Vote delegation failed

Reason: There's a reputation multiplier for delegated votes, it's set to 0.8. Unfortunately, that rounds to 0, and we're not allowed to add 0 reputation

Suggested fix: Add minimum 1 rep for adding into the votes table. 

Error replayed: 
 cleosm push action funds.seeds mimicvote '{
        "delegatee": "gradinagruiu",
        "delegator": "illumination",
        "scope": "milestone",
        "proposal_id": 244,
        "percentage_used": "0.49494949494949497",
        "option": "trust",
        "chunksize": 200
}' -p funds.seeds@active

Error 3050003: eosio_assert_message assertion failure
Error Details:
assertion failure with message: amount must be > 0

PR #414, opened 2021-11-20
Fix vote delegation

n13:
Fix for: https://github.com/JoinSEEDS/seeds-smart-contracts/issues/413

HC (Human cause): Setting parameters on mainnet could cause the rep value we add for delegators to be 0. addrep fails with an error when trying to addrep with 0. Unit tests modified the values and did not test the mainnet values, so unit tests did not cover this.

Fix: 
round the value, which with the current production settings means that delegators will get 1 rep point for entering into the table
Also, check for 0 - in case the settings change such that added rep would be 0, we don't call addrep.

### DIFF
--- a/src/seeds.dao.cpp
+++ b/src/seeds.dao.cpp
@@ -365,12 +365,14 @@ ACTION dao::erasepartpts (const uint64_t & active_proposals) {
 
   while (pitr != participants_t.end() && counter < batch_size) {
     if (pitr->count == active_proposals && pitr->nonneutral) {
-      send_inline_action(
-        permission_level(contracts::accounts, "addrep"_n),
-        contracts::accounts, 
-        "addrep"_n,
-        std::make_tuple(pitr->account, reward_points)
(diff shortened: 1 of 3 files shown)


src/seeds.dao.cpp
```diff
@@ -365,12 +365,14 @@ ACTION dao::erasepartpts (const uint64_t & active_proposals) {
 
   while (pitr != participants_t.end() && counter < batch_size) {
     if (pitr->count == active_proposals && pitr->nonneutral) {
-      send_inline_action(
-        permission_level(contracts::accounts, "addrep"_n),
-        contracts::accounts, 
-        "addrep"_n,
-        std::make_tuple(pitr->account, reward_points)
-      );
+      if (reward_points > 0) {
```
n13, line 368: this could be outside the look actually but under normal circumstances we expect the value to be > 0

IanMendozaJaimes (approved):

PR #415, opened 2021-11-29
Feature: Expiring policy

n13:
Expiring policy - set the expiration seconds in the Action call. 

Antonio requested this as to be able to implement authentication

## New Actions

```
createexp(name account, string backend_user_id, string device_id, string signature, string policy, uint64_t expiry_seconds);
```
This is the same as "create" only with an expiry in seconds added to it

see the last field - expiry_seconds

## New Tables

```
expiry
```

Contains these fields:
```
int id - id of the policy (foreign key into devicepolicy table)
int created_at - seconds since 1970 timestamp when the policy was created
int valid_until - seconds since 1970 timestamp when the policy expires 
 ```

### DIFF
--- a/src/seeds.policy.cpp
+++ b/src/seeds.policy.cpp
@@ -1,18 +1,28 @@
 #include <seeds.policy.hpp>
+#include <eosio/transaction.hpp>
 
-void policy::reset() {
+void policy::reset()
+{
   require_auth(_self);
 
   auto pitr = devicepolicy.begin();
-  while (pitr != devicepolicy.end()) {
+  while (pitr != devicepolicy.end())
+  {
     pitr = devicepolicy.erase(pitr);
+  }
+
+  auto eitr = expiry.begin();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/seeds.policy.cpp [hunk 1]
```diff
@@ -1,53 +1,124 @@
...
+void policy::update(uint64_t id, name account, string backend_user_id, string device_id, string signature, string policy)
+{
+
+  auto pitr = devicepolicy.find(id);
+
+  check(pitr != devicepolicy.end(), "policy with id not found: " + std::to_string(id));
+  check(pitr->account == account, "account cannot be changed: " + (pitr->account).to_string());
+
+  require_auth(pitr->account);
+
```
n13, line 81: I've changed update to actually check the creating account, not the account passed as parameter

I think that was a bug in the previous version

As the account passed as parameter could have been anyone, it was not checked against the creator account

PR #416, opened 2021-12-02
Fix cyclestats start_time and end_time using moon cycle times #412

### LINKED ISSUE #412: Fix cyclestats start_time and end_time using moon cycle times
## Why

Currently cycle stats start time is set to whenever onperiod is run
end_time is set to 29.5 days after

This is because in the past, we did not have actual moon cycle times on contract. We were estimating.

## Fix

On onPeriod, search for the closest new moon to the now() time, for the start
Then set end to the next New Moon after that

Using the moonphases tables in the cycle.seeds contract.

https://telos.bloks.io/account/cycle.seeds?loadContract=true&tab=Tables&account=cycle.seeds&scope=cycle.seeds&limit=100&table=moonphases

## How (suggestion)
1 - when onperiod is run, take the now() time, and subtract say 5 days in seconds from it (meaning it works as long as onperiod was run within 5 days of the new moon)
2 - use this value as lower bound search 
3 - iterate until the next new moon, use as start, 
4 - iterate more until second new moon, use as end

Cleos example to illustrate bounds search

```
cleosm --print-request get table cycle.seeds cycle.seeds moonphases --lower 982906450 --limit 1
REQUEST:
---------------------
POST /v1/chain/get_table_rows HTTP/1.0
Host: node.hypha.earth
content-length: 287
Accept: */*
Connection: close

{
  "json": true,
  "code": "cycle.seeds",
  "scope": "cycle.seeds",
  "table": "moonphases",
  "table_key": "",
  "lower_bound": "982906450",
  "upper_bound": "",
  "limit": 1,
  "key_type": "",
  "index_position": "",
  "encode_type": "dec",
  "reverse": false,
  "show_payer": false
}
---------------------
{
  "rows": [{
      "timestamp": 982916460,
      "time": "2001-02-23T08:21:00.000",
      "phase_name": "New Moon",
      "eclipse": ""
    }
  ],
  "more": true,
  "next_key": "983584980"
}

```

IanMendozaJaimes:
- Added moon phases table to contracts proposals and dao.
- Now cycle stats for both contracts (proposals and dao) uses moon phases as start and end dates

Fixes #412 

Contracts to be deployed:
- [ ] Proposals
- [ ] Dao

No permissions nor settings need to be updated

### DIFF
--- a/src/seeds.dao.cpp
+++ b/src/seeds.dao.cpp
@@ -950,6 +950,23 @@ void dao::send_mimic_delegatee_vote (const name & delegatee, const name & scope,
 
 }
 
+uint64_t dao::get_new_moon (uint64_t timestamp) {
+
+  moon_phases_tables moonphases_t(contracts::scheduler, contracts::scheduler.value);
+
+  auto mitr = moonphases_t.lower_bound(timestamp);
+
+  while (mitr != moonphases_t.end()) {
+    if (mitr->phase_name == "New Moon") {
+      return mitr->timestamp;
+    }
+    mitr++;
+  }
+
+  return 0;
+
+}
+
 void dao::init_cycle_new_stats () {
 
   cycle_tables cycle_t(get_self(), get_self().value);
@@ -957,10 +974,15 @@ void dao::init_cycle_new_stats () {
 
   cycle_stats_tables cyclestats_t(get_self(), get_self().value);
 
+  uint64_t now = eosio::current_time_point().sec_since_epoch() - (utils::seconds_per_day * 5);
+
+  uint64_t start_time = get_new_moon(now);
+  uint64_t end_time = get_new_moon(start_time + utils::seconds_per_day);
+
   cyclestats_t.emplace(_self, [&](auto & item){
     item.propcycle = c_t.propcycle;
-    item.start_time = c_t.t_onperiod;
-    item.end_time = c_t.t_onperiod + config_get("propcyclesec"_n);
+    item.start_time = start_time;
+    item.end_time = end_time;
     // item.num_proposals = 0;
     item.num_votes = 0;
     item.total_voice_cast = 0;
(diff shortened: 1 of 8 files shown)


src/seeds.dao.cpp
```diff
@@ -950,17 +950,39 @@ void dao::send_mimic_delegatee_vote (const name & delegatee, const name & scope,
...
 void dao::init_cycle_new_stats () {
 
   cycle_tables cycle_t(get_self(), get_self().value);
   cycle_table c_t = cycle_t.get();
 
   cycle_stats_tables cyclestats_t(get_self(), get_self().value);
 
+  uint64_t now = eosio::current_time_point().sec_since_epoch() - (utils::seconds_per_day * 5);
+
+  uint64_t start_time = get_new_moon(now);
```
n13, line 979: make it like this:

uint64_t start_time  = get_new_moon(start_time - utils::seconds_per_day);

This way we can be sure we get the current new moon... it may be a few seconds off...

src/seeds.dao.cpp
```diff
@@ -950,17 +950,39 @@ void dao::send_mimic_delegatee_vote (const name & delegatee, const name & scope,
...
+}
+
 void dao::init_cycle_new_stats () {
 
   cycle_tables cycle_t(get_self(), get_self().value);
   cycle_table c_t = cycle_t.get();
 
   cycle_stats_tables cyclestats_t(get_self(), get_self().value);
 
+  uint64_t now = eosio::current_time_point().sec_since_epoch() - (utils::seconds_per_day * 5);
```
n13, line 977: maybe this variable should be called now_minus_five_days 

I thought the code was buggy because it was called now, only later saw that it's correct ;)

src/seeds.proposals.cpp
```diff
@@ -914,13 +914,35 @@ void proposals::update_cycle() {
...
+  }
+
+  return 0;
+
+}
+
 void proposals::init_cycle_new_stats () {
   cycle_table c = cycle.get();
 
+  uint64_t now = eosio::current_time_point().sec_since_epoch() - (utils::seconds_per_day * 5);
```
n13, line 937: same - rename now

n13 (approved):
Approved with comments, variable name can be better

PR #420, opened 2022-01-02
Feature/master token list

chuck-h:
New "tokensmaster" contract; implements a master library of token metadata for tokens used in the Seeds ecosystems tools. Going forward this will be valuable as new tokens proliferate from regional DHOs and other Seeds-related organizations.

Tokens are added to the master library through a submit-approve sequence.

The intention is that Light Wallet will use this contract as a trustworthy source for `TokenModel` metadata. The contract is installed on the `modls.seeds` account. 

Issue #421


n13 (approved):

scripts/helper.js
```diff
@@ -229,6 +230,7 @@ const accountsMetadata = (network) => {
       pool: contract('pool.seeds', 'pool'),
       dao: contract('dao.seeds', 'dao'),
       startoken: contract('star.seeds', 'startoken'),
+      tokensmaster: contract('modls.seeds', 'tokensmaster'),
```
n13, line 233: What's modls stand for? Haha. If this is a generic contract, we may just use "tokenmasterx" or whatever. 

I mean if we expect people other than Seeds using it. 

tmastr.seeds ?  

tmgr.seeds ?

chuck-h: I made it `tmastr.seeds`

PR #460, opened 2022-04-20
Rainbow token contract update

chuck-h:
Bugfix, minor feature add (enable burning tokens without releasing staked resources), tests.


include/seeds.rainbows.hpp
```diff
@@ -215,11 +215,13 @@ using namespace eosio;
...
           * The opposite for issue action, if all validations succeed,
-          * it debits the statstable.supply amount. Any staked tokens are released from escrow in
-          * proportion to the quantity of tokens retired.
+          * it debits the statstable.supply amount. If `burn` flag is false,
+          * any staked tokens are released from escrow in proportion to the
+          * quantity of tokens retired.
           *
           * @param owner - the account containing tokens to retire,
           * @param quantity - the quantity of tokens to retire,
+          * @param burn - if true, staked tokens are left in escrow,
```
n13, line 224: Hmm... I don't understand what burn parameter does. You'd think it burns token if this is true. But.... that's not what it does?

chuck-h: The `retire` action on an ordinary token would just burn it, basically destroying value. But with "staked" or "backed" tokens, a `retire` action releases the staked value to the token holder and is more of a conversion of value. By setting the burn flag, a staked token is retired (value destroyed) but the staked value is not released. Burn flag is a NOP on a plain token with no backing.

n13: Maybe we can call it force_burn?

n13: or... do_not_redeem?
Hmm.

PR #466, opened 2022-05-25
Feature/master token list

chuck-h:
Added blacklist/whitelist tables, which gate the `submittoken` action. Also duplicated token symbols under different contracts cannot be submitted without an explicit whitelist entry.

More granularity to `accepttoken` authorization
1. support usecase-specific curator accounts, in addition to existing manager account
2. manager & curator accounts are empowered to transfer their authority to new accounts.


n13 (approved):
It's a lot of code - I guess at some point we should audit the whole codebase.

Issue #438, opened 2022-01-31
Citizen Path Change [Referendum #4]


n13:
## Implementation Spec

Change it so when 3 or more citizens vouch for a user, the user can instantly become a citizen

Remove the requirement for "inviting members"

## Original Spec
A. Path from Visitor to Citizen: becoming a citizen doesn’t require inviting members any longer but rather being vouched for by 3 Citizens during a Citizen Ceremony. 

This is a project in itself (SEEDS Citizenship), as it requires further design of the process for neighbourhubs, organisational implementation at the SEEDS level and technical implementation in the apps. 

Implementation Team for the SEEDS Citizenship project:
RE: → changes to the SEEDS Guide
MoM: → changes to the rituals and people process for finding Citizens to receive vouches
Hypha: → changes to the Smart contracts, Passport and Light Wallet

chuck-h:
https://discord.com/channels/722537361480613950/969298065490067466/974471310895882240

"[...] I therefore suggest that we take a "shortcut" implementation approach which provides a means to enroll new citizens through the new ceremony process. Specifically, let us create a new blockchain action which directly promotes an account to citizen status [...]."

n13:
Original change was not good enough, made another commit for this

n13:
Waiting for apps to update before deployment

PR #469, opened 2022-06-08
[GOVERNANCE Referendum #4] citizenship ceremony

n13:
Fix for #438 

Additional change was required, we no longer check for the rep _score_ for citizens

The reason is that the score is calculated depending on all other members, meaning it could prevent the citizenship ceremony from working in rare cases where 50 reputation points - which we get from resident status - is not enough to ensure a score of 50

This edge case was always in there, but since there's no way to directly affect the score, it was kind of a bug.


n13:
Merged and deployed

PR #471, opened 2022-06-10
Master Token List -- install JSON schema in on-chain table

chuck-h:
This embeds the schema so that apps throughout the ecosystem (e.g. Light Wallet, Rainbow token manager, etc.) can access it.


chuck-h:
Suggest during deployment we create a dedicated master token list management account, e.g. `tokmgr.seeds` and specify that as "manager" when the contract is deployed to mainnet and the one-time `init` action is called. Owner permission of this account would be cg.seeds, active permission could be assigned to whoever is holding the DHO role of manager (could initially be Nik). Ref https://discord.com/channels/722466006206906428/971085405162209320/984601564960395335

Would love to see this deployed and preloaded with HYPHA, HUSD, TELOS, etc for testing of the LW side in LW PR 1880

chuck-h:
> Why do we need a schema?
> 
> It seems to me a wrong approach to store a schema on chain?

Yeah, maybe not the most elegant design decision for having an "authoritative location" for a schema. We'll see if it's ever useful.

PR #475, opened 2022-11-02
Bulk onboarding script update for invite manager

chuck-h:
The existing onboarding script includes a `bulk_invite` command, which generates a quantity of invites, places them on chain, and saves a file `secrets_nnn.csv` containing the secrets and hashes.

This PR adds an additional output file `invitelist_nnn.txt`  containing lines of the form
```
https://joinseeds.app.link/accept-invite?invite-secret=354a364a624d46784c546d645a6855704755613767437167517674705a684c47
```
This is consistent with the invite "export" function of the Seeds Global Passport web application, and can be imported to the invite manager at https://invites.joinseeds.earth/login

image

### DIFF
--- a/scripts/onboarding_script.js
+++ b/scripts/onboarding_script.js
@@ -144,6 +144,7 @@ const bulk_invite = async (sponsor, referrer, num, totalAmount) => {
 
     totalAmount = parseInt(totalAmount)
     var secrets = "Secret,Hash,Seeds (total)\n"
+    var invitelist = ""
     const fileName = 'secrets_'+num+'.csv'
     var log = []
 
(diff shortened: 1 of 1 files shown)


scripts/onboarding_script.js
```diff
@@ -309,7 +311,7 @@ const createCPUAction = (sponsor) => {
         authorization: [
             {
                 actor: harvest,
-                permission: 'active',
+                permission: 'payforcpu',
```
n13, line 314: That seems wrong - should be active.

chuck-h: The `payforcpu` permission exists for the limited purpose of executing the `payforcpu` action (see key table below). While `active` is also capable of executing this, it is a more powerful authorization and IMHO it is good practice to use the least powerful auth that does the job.

In this particular case, I as a user do not hold the harvst.seeds active key so I could not execute the script. On the other hand the payforcpu key is not strictly controlled (I scraped it from the LW repo). In the end I think this level of auth makes more sense.
image

n13: Oh - I see so that was a bug in the script.

n13:
> leave active permission

Actually active permission there was a bug - payforcpu is correct for the payforcpu action.

n13 (approved):

n13 (approved):

n13 (approved):

n13:
@chuck-h please merge

